PR #207, opened 2022-02-07
Adding Docker image kaotoio/frontend and job to build it

Delawen:
The job will generate a nightly "latest" and also is supposed to generate a stable version for each tag we push to the repository.

It is based on the node 16 image instead of node 17 because the node 17 has some compatibility issue.


Delawen:
Missing from here: how to make yarn listen to anything not localhost.

kahboom (approved):
Looks good to me 👍

PR #228, opened 2022-02-22
Rename REACT_APP_API_URL to KAOTO_API

mmelko:


kahboom (approved):
thanks so much @mmelko !

PR #230, opened 2022-02-25
Contextual Mini Catalog

### LINKED ISSUE #216: Contextual Mini Catalog
Add support for a "mini catalog" that loads a list of steps. The idea is that when a user wants to add a new step to an existing step, they will click a + button that loads a searchable mini catalog of steps compatible with it. The user can then select the step, which will automatically be appended to the existing step.

### Steps
- [x] [Update apiService to use new endpoint](https://kaotoio.github.io/kaoto-backend/api/index.html#api-StepResource-stepGet)
- [x] Add component for Contextual Mini Catalog
- [x] Add handler for selecting steps
- [x] Verify validation is in tact, add tests
- [x] Add search

Depends on https://github.com/KaotoIO/kaoto-backend/issues/42 and either #218 or #219 for the final implementation, but development can start before then.

kahboom:
Adds support for the Contextual Mini Catalog, and its corresponding popover. resolves #216 

2022-02-24 at 16 37 51 - Append Step

cc @Delawen

### DIFF
--- a/src/api/StepsAndViewsProvider.tsx
+++ b/src/api/StepsAndViewsProvider.tsx
@@ -7,6 +7,7 @@ interface IStepsAndViewsProvider {
 }
 
 type StepsAndViewsAction =
+  | { type: 'ADD_STEP'; payload: { newStep: IStepProps } }
   | { type: 'DELETE_STEP'; payload: { index: number } }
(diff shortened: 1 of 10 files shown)


kahboom:
@Delawen - I've merged it in without waiting for the endpoint. I can add some validation in, in a follow-up PR, so that only relevant steps will be shown or selectable, but if the endpoint does become available just let me know.

PR #258, opened 2022-03-17
reduce docker image size

evanshortiss:
This reduces the image size from over 2.5GB to 174MB. I verified that the UI loads in my browser using the following commands, and opening http://localhost:8080. 

```
podman build . -t kaoto
podman run --rm -p 8080:80 kaoto:latest
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+node_modules
+dist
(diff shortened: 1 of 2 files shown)


evanshortiss:
BTW, adding an `nginx.conf` file might be worth looking into for routing, 404s, etc but this is a good start I think.

kahboom (approved):
you're a legend @evanshortiss thanks so much for your contribution!  🥳 🥇

PR #272, opened 2022-03-29
Make possible to set KAOTO_API as a parameter for the docker image

mmelko:


src/api/request.ts
```diff
@@ -31,7 +31,13 @@ export interface IFetch {
   stringifyBody?: boolean;
 }
 
-const apiURL = process.env.KAOTO_API;
+let apiURL = process.env.KAOTO_API;
+console.log(process.env.NODE_ENV);
+
+if (process.env.NODE_ENV == 'production' && undefined != window.KAOTO_API) {
```
kahboom, line 37: Sorry what I meant was changing this line to this: `if (process.env.NODE_ENV == 'production' &&  window.KAOTO_API) {` because I think it would check that `window.KAOTO_API` would not be null or undefined, you don't have to explicitly check that it's not equal to `undefined`. See here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_NOT#syntax

kahboom (changes requested):

kahboom:
Also, @mmelko it seems like unit tests are failing because it's not finding `KAOTO_API` in `Window`, so it might be that you need to add it to the `src/types.d.ts` file.

kahboom (approved):
Thanks @mmelko ! 🎉

Issue #274, opened 2022-03-30
failed to decode downloaded font webpack


kahboom:
Getting this warning message in the browser console now that I'm pulling in a Google font to fix the font issue:

`failed to decode downloaded font webpack`

Likely related to having to use Webpack's new asset management features since version 5.x.

PR #277, opened 2022-04-02
Use assets modules instead of deprecated loaders

Pompette:
#274 

Current loaders where deprecated since webpack 5, changed it to [assets modules](https://webpack.js.org/guides/asset-modules/#resource-assets)


kahboom (approved):
thanks for your contribution @Pompette , I really appreciate it! ❤️

Issue #264, opened 2022-03-23
Uploading YAML doesn't update Visualization


kahboom:
To reproduce:
- Select to upload a YAML file from the Code Editor
- YAML file is now shown in the Code Editor, but there are no changes to the Visualization unless you change something in the code editor.

kahboom:
closed via #278

PR #278, opened 2022-04-03
Fix upload YAML doesn't upload visualization

Pompette:
#264 

Hello, here is a proposal for fixing YAML doesn't upload visualization, I've change the way of listening for change. Now it's using the monaco editor "onDidChangeModelContent" event.


kahboom (approved):
Just tested the branch locally and it's perfect, thanks so much @Pompette !

PR #279, opened 2022-04-03
Add dark mode

Pompette:
#265 

darkMode

Add dark mode and save theme preferences to localStorage

PR #286, opened 2022-04-04
Change way of creating react node for header tools

Pompette:
#279 

@kahboom well see, thanks for your return, I've made the modification 👍🏽


kahboom (approved):
thanks @Pompette ! 🥇

PR #288, opened 2022-04-05
Improve e2e tests

Delawen:
Adding backend service.

Separate cypress install and run to add multiple browsers.

DO NOT MERGE UNTIL TESTS ARE GREEN!!


Delawen:
@kahboom this is ready, except for two steps that fail (outside this PR scope)

PR #395, opened 2022-05-16
Migrate to unprivileged nginx image to work also on restricted environments

mmelko:
Migrate to unprivileged nginx image to work also on restricted environments
Add docker-compose config to run with docker-compose.


Dockerfile
```diff
@@ -13,13 +13,13 @@ COPY . .
...
 
-FROM nginx:latest
+FROM nginxinc/nginx-unprivileged
 
 COPY --from=appbuild /app/dist/* /usr/share/nginx/html/
-COPY nginx/start_nginx.sh .
+COPY nginx/nginx.conf /etc/nginx/nginx.conf
 
-EXPOSE 80
+EXPOSE 8080
```
Delawen, line 21: But we want it to be on 80, no?

PR #471, opened 2022-06-03
[Editor] add Undo/Redo buttons

### LINKED ISSUE #269: Undo/Redo for Code Editor
It would be useful for users to be able to undo/redo changes they make to the Code Editor.

mmelko:
close #269


mmelko:
@kahboom I'm not sure if calling undo/redo directly on the Model is the best solution, or I should rather  register an action for the editor and call it with `editor.trigger('','undo')`.

PR #478, opened 2022-06-06
issues with cypress github actions

kahboom:
Changes:

- Downgrade from Yarn berry to Yarn 1.x. Remove all relevant files, including the giant cache dir.
- Update the Dockerfile to reflect Yarn changes.
- Remove Yarn berry workarounds for Cypress to work with it (i.e. `command-prefix`)
- Remove problematic prebuild step causing the Docker builds to fail.


Delawen:
Please, remove the dependabot fix job too

Issue #511, opened 2022-06-15
[e2e] Make Kaoto location configurable


mmelko:
Currently in e2e tests the address is hardcoded:  : `cy.visit('http://localhost:1337');` .
It would be nice to refactor the tests and make the target URL configurable. This would allow to run the tests also against docker image or kaoto running remotely on some cluster.

Delawen:
@miram13 can you check @mmelko comment?

miram13:
Ya Sure

PR #514, opened 2022-06-16
Cypress test to Make Kaoto location configurable

miram13:
Which issue(s) this PR fixes: #511


mmelko:
I'm not sure if this will help if address is still hardcoded in the tests, for example: 
https://github.com/miram13/kaoto-ui/blob/8aa54121e35e297465e8e22abfbb2ba7090c2ad1/cypress/integration/6-step_integration/step_integration.spec.js#L5

PR #533, opened 2022-06-25
[DRAFT] layout changes

kahboom:
- Changed floating toolbar -> PF toolbar, moved everything there
- Removed sidebar
- Add list of deployments
- Catalog is now embedded instead of an overlay
- Add deployment status
- Hide code by default
- Update tests

This is kind of a part 1 out of 2 for the layout changes, just trying to keep the PRs somewhat reasonable.


mmelko:
just a cosmetic one.. but in the code there is 'integration' mentioned quite few times.  I'm just wondering if we want to be multiDSL if we shouldn't find some more general term for it :)

mmelko (approved):

PR #557, opened 2022-07-05
Use main image instead of nightly

Delawen:
Sometimes nightly arrives too late.


Delawen:
Of course it fails, you impatient urchin, the new backend image that is not broken is not ready yet. Give me a break!

PR #568, opened 2022-07-08
Add colors to warning icons

mmelko:
@kahboom fyi 

Due to connections problem I couldn't verify with deployment list


kahboom (approved):
this is perfect, thanks @mmelko !

Issue #608, opened 2022-07-26
Persist Steps: Between Changes


kahboom:
Belongs to Epic: #607 

## Problem
When you edit an integration, every time you make a change (i.e. add a step, remove a step), the steps are regenerated automatically. The nodes in the Visualization are then created in a linear fashion. If users have moved around the positions, those positions are then lost.

## Proposition
Merge position data from the UI with the new positions. If a step is removed, we can still use the same original positions (length of the array permitting).

PR #617, opened 2022-08-10
#608 persist steps between changes

kahboom:
node persistence changes:
- steps to take slot positions even on change
- delete node manually before step gets deleted

step extension API changes:
- move extension API outside of `StepViews` component
- provide initial values
- update types to remove unnecessary values (`connectorType`)

2022-08-10 at 17 37 15

implements #608


src/components/StepViews.tsx
```diff
@@ -43,6 +43,36 @@ export interface IStepViewsProps {
...
+    onKaotoButtonClicked,
+    saveConfig,
+    startDeployment: startKaotoDeployment,
+    step,
+    stepInitialValues,
+    stopDeployment: stopKaotoDeployment,
+    updateStep: (newStep: IStepProps) => {
+      // update state of step
+      replaceStep(newStep, index);
+    },
```
mmelko, line 70: I'd personally restrict the ap only to things required by stepExtensions.. which is probably just updateStep, and notifyKaoto. Also I'd rename the api to 'stepExtensionApi' and this save maybe for the general extension in future, maybe? 🤷‍♂️

PR #618, opened 2022-08-16
uxd initial review changes

kahboom:
This PR addresses several points mentioned in UXD's first review of Kaoto:

Screen Shot 2022-08-16 at 12 21 45 pm
1. Swap out tiles for burger menu icon on Toolbar
2. Move Code Editor & Catalog icons closer together
3. Move Delete button (moved into dropdown)
4. Initial ADD STEP placeholder should be greyed out
5. Move mini map & controls together, but, ultimately, decided to hide the mini map for now (will create a separate issue to enhance it) because it's quite useless as you can't pan around yet
6. Made changes to attribution as requested
7. Based on @lhein 's feedback, changed it from a Play button to `Deploy`

Screen Shot 2022-08-16 at 11 34 18 am

- Re-added header with helper text in Step Catalog
- Re-added close icon
- Add arrowhead to lines (small, you can barely see it, but it's there), also made all steps blue (we can change it to be animated on deployment or red/orange if not configured properly)

Screen Shot 2022-08-16 at 12 41 38 pm

Screen Shot 2022-08-16 at 12 22 00 pm

- Moved Light/Dark mode into separate `Appearance` settings
- Grouped together items in dropdown


lhein:
Great. I still struggle a bit with how the settings work. In my desktop tooling world the settings menu would contain things like modifying the appearance of the UI but it seems in Kaoto Settings is more related to the project / integration. Maybe something to check with UX would be to have a "Project Settings" icon near the integration name and then use the right side settings menu for global settings only (which would include appearance settings)?

kahboom:
>Great. I still struggle a bit with how the settings work. In my desktop tooling world the settings menu would contain things like modifying the appearance of the UI but it seems in Kaoto Settings is more related to the project / integration. Maybe something to check with UX would be to have a "Project Settings" icon near the integration name and then use the right side settings menu for global settings only (which would include appearance settings)?

@lhein I'm not sure if you were following the discussion on the UX review slides, but Missy also consulted with the UXD team about the naming, and, for me, it came down to this:
>I reached out to the UX content team regarding this menu item they think Settings. Configuration or Preferences would work here so we can keep as is. We could move the dark/light mode in this dialog for now (until there are more "user" settings) and add subheadings to separate the WORKFLOW from the APPEARANCE.

So, to avoid scrolling in the settings with several sections, I decided to separate them into two modals. Alternatively we could 1) add a cog icon that opens all settings in one modal, like Milanote does, but we'd probably have to move the Delete button back to the toolbar, or 2) maybe add it in the left-hand dropdown, but that also doesn't feel super intuitive to me.

Personally I've never seen configuration settings accessed near something like the name, or in the middle of the toolbar, but maybe I'm not thinking of the right thing.

PR #632, opened 2022-08-26
catalog seearch utilization and gitignore config

erayerdem:


kahboom:
thanks for your contribution @erayerdem ! For the type declarations I was about to move them to be copied to `dist/@kaoto/index.d.ts` and it was only going to be for the `IStepExtensionApi` interface (renamed to `IKaotoApi` in this PR: https://github.com/KaotoIO/kaoto-ui/pull/626), but if you think it would be useful to have all the type declarations in the repo we can keep them? let me know what you think. :)

erayerdem:
.gitignore no have dist folder therefore a bit consufed my commit I will rearrange pr . u are true 👍🏿

PR #647, opened 2022-08-31
feature(docker): Adding a health check before pushing

Delawen:


Delawen:
You can't have a health check without a backend running...

kahboom (approved):

Delawen:
This can't be done unless we can run the frontend docker image outside docker-compose.

PR #667, opened 2022-09-01
chore(build): Add @module-federation/typescript plugin

igarashitm:
Submitting a draft PR to share my quick try - I found that once the root component `App` is exposed, yarn start works just fine. But only `App.d.ts` is generated in `dist/`. I tried to expose `src/@kaoto/index.d.ts` members but no luck so far.

```
ERROR in ./src/@kaoto/index.d.ts
Module build failed (from ./node_modules/ts-loader/index.js):
Error: Debug Failure. Output generation failed
    at Object.transpileModule (/home/tomo/workspace/Kaoto/kaoto-ui/node_modules/typescript/lib/typescript.js:138944:29)
    at getTranspilationEmit (/home/tomo/workspace/Kaoto/kaoto-ui/node_modules/ts-loader/dist/index.js:397:74)
    at successLoader (/home/tomo/workspace/Kaoto/kaoto-ui/node_modules/ts-loader/dist/index.js:37:11)
    at Object.loader (/home/tomo/workspace/Kaoto/kaoto-ui/node_modules/ts-loader/dist/index.js:22:5)
```


igarashitm:
OK, so the filename `index.d.ts` was the problem. Once it's renamed to `index.ts`, the plugin generated `index.d.ts` under `dist/`. Tests are still failing though.

igarashitm:
Oh OK, if my change is not causing e2e test failure, otherwise the plugin seems to be working as expected - changing to "ready for review"

igarashitm:
ugh good catch, I almost forgot about that - it smells like a bug in the plugin, but if I create a `dist/@mf-typescript` directory, then the error went away.

igarashitm:
For now, we might need to ensure the directory exists or create?

kahboom:
after I pulled in the dependency upgrades it seemed to go away on its own. Unfortunately, checking in the `dist` dir is not really something I'd want to do, nor to require it from users. But I'll check to see if it happens again.

kahboom (approved):
thanks very much @igarashitm !

PR #723, opened 2022-09-19
feature(test): Adding coverage and codecov reporting

Delawen:
Let's see if it works.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -1,4 +1,4 @@
-name: Unit Tests (Jest)
+name: 🧪 Unit Tests (Jest) 🧪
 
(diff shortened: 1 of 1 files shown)


Delawen:
SonarCloud, are you jealous because I added Codecov? My commit does not have any duplications.

kahboom:
@Delawen why did we add another check when it's failing?

Delawen:
You mean sonarcloud? 

Well, we can tweak the levels at which we want it to fail. But it is better to know it than to have no checks.

Delawen:
Btw: sonarcloud quality gate only considers the code from the last few days. This one is considering all the code because it was the first one. That's why it fails.

PR #727, opened 2022-09-23
Fetch default namespace from the backend

### LINKED ISSUE #726: Fetch deafult namespace from the backend instead of using hardcoded "default"
This will be much more convenient for using in the openshift/kubernetes with privileges limited only to particular namespace.

mmelko:
close #726 
fetching the default namespace is done in useEffects KaotoModal.  Also fetching of deployments were done even before so I added checks to fetch deployments only if there is namespace set. 

There is probably a better way how to do this but I'm not sure yet.

### DIFF
--- a/src/components/DeploymentsModal.tsx
+++ b/src/components/DeploymentsModal.tsx
@@ -56,25 +56,29 @@ export const DeploymentsModal = ({ handleCloseModal, isModalOpen }: IDeployments
 
   useEffect(() => {
     // fetch deployments
-    fetchDeployments('no-cache', settings.namespace)
-      .then((output) => {
-        setDeployments(output);
-      })
-      .catch((e) => {
-        throw Error(e);
-      });
+    if (settings.namespace !== '') {
+      fetchDeployments('no-cache', settings.namespace)
+          .then((output) => {
+            setDeployments(output);
+          })
+          .catch((e) => {
+            throw Error(e);
+          });
+    }
   }, [settings.namespace]);
 
   // on changes to deployment, re-fetch list of deployments
   useEffect(() => {
     // fetch deployments
-    fetchDeployments('no-cache', settings.namespace)
-      .then((output) => {
-        setDeployments(output);
-      })
-      .catch((e) => {
-        throw Error(e);
-      });
+    if (settings.namespace !== '') {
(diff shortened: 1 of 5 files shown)


src/components/DeploymentsModal.tsx
```diff
@@ -56,25 +56,29 @@ export const DeploymentsModal = ({ handleCloseModal, isModalOpen }: IDeployments
...
     // fetch deployments
-    fetchDeployments('no-cache', settings.namespace)
-      .then((output) => {
-        setDeployments(output);
-      })
-      .catch((e) => {
-        throw Error(e);
-      });
+    if(settings.namespace !== '') {
+      console.log('fetcging namespace',settings.namespace)
```
Delawen, line 60: A bit nitpicky but there is a typo here :)

src/components/DeploymentsModal.tsx
```diff
@@ -56,25 +56,29 @@ export const DeploymentsModal = ({ handleCloseModal, isModalOpen }: IDeployments
...
+            throw Error(e);
+          });
+    }
   }, [settings.namespace]);
 
   // on changes to deployment, re-fetch list of deployments
   useEffect(() => {
     // fetch deployments
+    if(settings.namespace !== '') {
     fetchDeployments('no-cache', settings.namespace)
```
Delawen, line 75: More nitpicky: shouldn't this have some indentation?

Delawen:
LGTM

src/components/DeploymentsModal.tsx
```diff
@@ -56,25 +56,30 @@ export const DeploymentsModal = ({ handleCloseModal, isModalOpen }: IDeployments
...
     // fetch deployments
-    fetchDeployments('no-cache', settings.namespace)
-      .then((output) => {
-        setDeployments(output);
-      })
-      .catch((e) => {
-        throw Error(e);
-      });
+    if (settings.namespace !== '') {
+      console.log('fetching namespace', settings.namespace)
```
kahboom, line 60: @mmelko - I think you left this here by mistake maybe?

mmelko: yep, thanks! 
It should be removed now

kahboom (approved):

Issue #675, opened 2022-09-02
[UX] Provide a "Delete" button for steps on canvas


lhein:
**Is your feature request related to a problem? Please describe.**
Adding new figures / steps to the canvas is really easy and can be done directly on the canvas which is quite convenient. Unfortunately to remove a figure from the canvas you have to click it to open the right side configuration panel and to click the "Delete" button there. 

**Describe the solution you'd like**
Can we add a "-" button to the figure analogue to the "+" one?

**Additional context**
![grafik](https://user-images.githubusercontent.com/435853/188118102-ac093cb1-e3a1-4653-8e8a-f1430c2979f2.png)

kahboom:
@miram13 - As we discussed on our call earlier today, this is how I would approach this issue, but feel free to do it differently:

### Functionality
We already have a function to delete steps that can be reused (the same one we call when deleting a step from the Step Detail View) [here](https://github.com/KaotoIO/kaoto-ui/blob/main/src/components/Visualization.tsx#L174). However, for it to be reused we need to move `setIsPanelExpanded(false);` out of there (because we don't need to close the panel if we delete the step directly from the canvas).

1. I suggest that instead of calling `handleDeleteStep` directly as it does [here](https://github.com/KaotoIO/kaoto-ui/blob/main/src/components/Visualization.tsx#L267), you instead convert it to its own function like this:

```
() => {
  handleDeleteStep();
  setIsPanelExpanded(false);
}
```

2. Pass the `handleDeleteStep` method to the `VisualizationStep` component by adding a reference to it [here](https://github.com/KaotoIO/kaoto-ui/blob/main/src/components/Visualization.tsx#L137), just like we do with `step`.
3. Your IDE will complain because `IVizStepPropsNode` is an [interface](https://www.typescriptlang.org/docs/handbook/typescript-tooling-in-5-minutes.html#interfaces) in TypeScript, almost like a blueprint for what properties it allows, and it doesn't understand what is `handleDeleteStep`. You will have to edit that interface to also allow a function called `handleDeleteStep` or whatever you choose to call it. This is done from [here](https://github.com/KaotoIO/kaoto-ui/blob/main/src/%40kaoto/index.ts#L152). It could look something like this: `handleDeleteStep: () => void`.
[...]

PR #746, opened 2022-09-29
feat(viz): Provided a delete button for steps on canvas

miram13:
This PR adds a delete button to steps on the Visualization canvas on hover, and removes the Delete button from Step Details.

Screen Shot 2022-10-19 at 4 15 53 pm

Which issue(s) this PR fixes:  #675

### DIFF
--- a/src/components/PlusButtonEdge.css
+++ b/src/components/PlusButtonEdge.css
@@ -1,4 +1,4 @@
-.plusButton {
+.trashButton, .plusButton {
     width: 20px;
     height: 20px;
     background: var(--pf-global--Color--light-300);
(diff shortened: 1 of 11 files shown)


kahboom:
@miram13 a few points here:

- Delete button from Step Detail now doesn't work, but you're in luck because I think I will be removing it once your PR is merged. 😂 
- Before merging this PR I would like to see this delete button hidden until the user hovers over the step, so it doesn't look so cluttered.
- If we can move the button to be on top or somewhere more out of the way (like this maybe?) that would be great. This will probably involve CSS. I'll check where this needs to be done.

src/components/PlusButtonEdge.css
```diff
@@ -25,3 +25,31 @@
...
+
+.minusButton-foreignObject span {
+    background: transparent;
+    width: 40px;
+    height: 40px;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    min-height: 40px;
+}
```
kahboom, line 55: Don't forget to add the extra space here :D

kahboom (changes requested):
@miram13 the remaining steps I think are all just CSS, really:
- add a `className` to the delete button
- in the CSS, you can show/hide it with `.className:hover {}`
- make the button smaller
- position the button a bit higher and maybe top-right or top-left of the circle, similar to the icons you see in the example below from OpenShift topology view
- BONUS: in the PR it's always good to provide a screenshot of your changes, for visual feedback, if it's a visual change

Personally I don't think we need to make the button red, I think it can stay the same color, but we will need some modal confirmation as well to make sure users don't delete their steps by mistake. This can be done in a follow-up PR.

Screen Shot 2022-10-12 at 1 28 56 pm

kahboom (approved):

PR #750, opened 2022-09-30
Adds integration with Kogito functionalities to be able to run in a VSCode extension

thiagoelg:


kahboom:
@thiagoelg did you want for me to review this already or wait until the other tasks are checked off? and thanks btw! :)

thiagoelg:
@kahboom feel free to review :)

But right now there seems to be a problem with the new version of React Flow that you might be able to help me with.
Running Kaoto UI by itself works fine, but running it inside VSCode makes it so the canvas is not interactable, and clicking and dragging are not responsive.

Code_cX8oWALfQA
(The gif doesn't show clicks, but I was trying to click and drag)

This seems like a CSS problem, but I wasn't able to find the source of the problem so far.

thiagoelg:
Ok, I think I fixed it... seems like importing CSS from inside a local CSS file works better for the VSCode extension (something related to how webpack resolve these files)

kahboom:
Glad to hear that @thiagoelg , if you run into any other problems let me know.

package.json
```diff
@@ -43,13 +57,15 @@
     "react-router-dom": "5.3.4",
     "react-router-last-location": "2.0.1",
     "reactflow": "^11.1.0",
+    "simple-zustand-devtools": "^1.1.0",
```
kahboom, line 60: @thiagoelg was it failing when this lib was a dev dependency? Maybe I can check the environment before trying to load this (which I believe is done at the bottom of the `integrationJsonStore.tsx` file).

src/components/import.ts
```diff
@@ -3,6 +3,7 @@ export async function dynamicImport(
   module: string | undefined,
   url: string | undefined
 ) {
+  // @ts-ignore
```
kahboom, line 6: was this causing issues or did you forget to remove it?

thiagoelg: It was causing problems 😕:

```
src/components/import.ts:6:9 - error TS2304: Cannot find name '__webpack_init_sharing__'.

6   await __webpack_init_sharing__('default');
          ~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error.

error Command failed with exit code 2.
```

kahboom:
@thiagoelg this looks great, I reviewed it but I'm wondering, what is the easiest way to test it?

thiagoelg:
> @thiagoelg this looks great, I reviewed it but I'm wondering, what is the easiest way to test it?

Probably via the VSCode extension, there are some instructions in this PR: https://github.com/KaotoIO/vscode-kaoto/pull/30

Testing VSCode extensions is a bit tricky, let me know if something doesn't work or if you need any help. I could probably send a built .vsix file too if necessary

package.json
thiagoelg: It failed when being used in web tools, since just importing it was causing issues. (also, `process.env` is directly evaluated for ts libraries 😕)

package.json
```diff
@@ -1,17 +1,24 @@
...
   "scripts": {
     "start": "webpack serve --mode development --hot --progress --config webpack.dev.js ",
     "prod": "webpack serve --mode production --port 8080 --allowed-hosts all --progress --config webpack.prod.js ",
     "build": "webpack --mode production --config webpack.prod.js",
+    "build:dev": "webpack --mode production --config webpack.dev.js",
+    "build:lib": "rimraf dist/lib && yarn copy:css && yarn copy:images && tsc -p tsconfig.lib.json && node ./scripts/fixTsPathsAlias.js",
+    "build:lib:watch": "tsc-watch --noClear -p tsconfig.lib.json",
+    "copy:css": "copyfiles -u 1 \"src/**/*.{sass,scss,css}\" dist/lib/",
+    "copy:images": "cpr ./src/assets ./dist/lib/assets",
+    "build-storybook": "build-storybook -s public",
```
kahboom, line 20: We removed Storybook and Chromatic, so these can be removed from here too :)

kahboom (changes requested):
@thiagoelg - Tiny change request, after that it'd be good to squash the commits. Then we're good to go!

kahboom (approved):
Great work @thiagoelg  🥳 🎉

PR #752, opened 2022-10-01
Remove default fetchCatalogSteps call from Catalog component

### LINKED ISSUE #692: Duplicated call to /v1/steps endpoint
**Describe the bug**
When opening the step catalog, there is a duplicated call to the `/v1/steps` endpoint.

**To Reproduce**
1. Go to kaoto 
2. Open Developer Tools
3. Click on step catalog
4. See the network tab

image

**Expected behavior**
I guess only one call should be needed. Both have the same parameters.

**Screenshots**
image

aaj013:
Closes #692
Remove default fetchCatalogSteps call from Catalog component to avoid duplicate steps get request


kahboom (approved):
thank you very much for your contribution @aaj013  🙂

PR #761, opened 2022-10-04
Reflect changes in backend: fetch step parameters individually

### LINKED ISSUE #770: Reflect steps endpoint changes in the backend
List of steps wont contain step parameters after https://github.com/KaotoIO/kaoto-backend/pull/174,  so this needs to be reflected in the UI. For retrieving the parameters `/steps/{id}` endpoint can be used

mmelko:
close #770

### DIFF
--- a/src/api/apiService.ts
+++ b/src/api/apiService.ts
@@ -36,6 +36,22 @@ export async function fetchDefaultNamespace() {
   }
 }
 
+export async function fetchStepDetails(id?: string, namespace?: string) {
+  try {
+    const resp = await request.get({
+      endpoint: `${apiVersion}/steps/id/${id}`,
+      queryParams: {
+        namespace: namespace ?? 'default',
+      },
+    });
+
+    return await resp.json();
+  } catch (err) {
+    console.error(err);
+    return err;
+  }
+}
+
 /**
  * Fetch all Catalog Steps, optionally with specified query params
  * @param queryParams
--- a/src/components/VisualizationStep.tsx
+++ b/src/components/VisualizationStep.tsx
@@ -1,4 +1,5 @@
 import './Visualization.css';
+import { fetchStepDetails } from '@kaoto/api';
 import { MiniCatalog } from '@kaoto/components';
 import {
   appendableStepTypes,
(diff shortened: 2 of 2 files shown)


src/api/apiService.ts
```diff
@@ -36,6 +36,23 @@ export async function fetchDefaultNamespace() {
   }
 }
 
+export async function fetchStepDetails(id: string | undefined, namespace?: string) {
```
kahboom, line 39: for this we can do `id?: string` instead of `string | undefined`, just like with `namespace`

src/components/VisualizationStep.tsx
```diff
@@ -15,6 +15,7 @@ import { AlertVariant, Popover } from '@patternfly/react-core';
 import { CubesIcon, PlusIcon } from '@patternfly/react-icons';
 import { useAlert } from '@rhoas/app-services-ui-shared';
 import { Handle, Node, NodeProps, Position, useNodes } from 'reactflow';
+import {fetchStepDetails} from "@kaoto/api";
```
kahboom, line 18: tiny comment here but the spacing! to keep it consistent. also same comment about the double quotes vs single. I will create an issue to add prettier to part of the workflow so we don't have to worry about this.

src/components/VisualizationStep.tsx
```diff
@@ -33,7 +34,13 @@ const VisualizationStep = ({ data }: NodeProps<IVizStepNodeData>) => {
 
   const { addAlert } = useAlert() || {};
 
-  const onMiniCatalogClickAdd = (selectedStep: IStepProps) => appendStep(selectedStep);
+  const onMiniCatalogClickAdd = (selectedStep: IStepProps) => {
```
kahboom, line 37: This might end up conflicting with Iram's PR, I will have to check

kahboom (changes requested):
Some nitpicks and small feedback :)

src/components/VisualizationStep.tsx
mmelko: well that's about merge order .. if you merge this later I will rebase and fix conflicts, otherwise it's up to Mohammadi :)

src/components/VisualizationStep.tsx
mmelko: yeah, thanks! formatting the code from my IDE formats everything even your code so I'm doing this manually.

kahboom (approved):
looks great, thank you!

PR #860, opened 2022-10-25
Update name when changed from the internal source code editor or vscode

mmelko:


Delawen (approved):

Delawen:
LGTM, should we merge?

src/components/SourceCodeEditor.tsx
```diff
@@ -43,6 +43,10 @@ const SourceCodeEditor = (props: ISourceCodeEditor) => {
       fetchIntegrationJson(incomingData, settings.dsl.name)
         .then((res: IIntegration) => {
           let tmpInt = res;
+          if (res.metadata.name != '') {
+            settings.name = res.metadata.name;
+            setName(res.metadata.name);
```
kahboom, line 48: Unrelated to your PR but I wonder why we have a separate `setName` method in `useSettingsStore`? 🤔  Maybe we can just use `setSettings({ name: res.metadata.name });` instead and remove the `setName`? Would you be willing to change that?

mmelko: I don't mind to change that. Do you want me to do it within this PR ?

kahboom (approved):

PR #868, opened 2022-10-28
Add a waiting screen instead of blank page

### LINKED ISSUE #826: Provide a waiting screen in case backend is not available/ready

In case, the kaoto-backend is not ready, currently a "loading" icon is briefly displayed then there is a blank page. it would be convenient to provide more information to end-user and if possible to wait few seconds for the backend to get started.

I imagine it can be done in different iteration:
- a simple error message mentioning backend cannot be reach and to try refresh browser later
- a simple error message mentioning backend cannot be reach and a button to refresh (useful in VS Code integration to avoid end-user to close and reopen the editor)
- an error message mentioning backend cannot be reach, a button to refresh and a spinning for few seconds with a process retrying on the background during that time to connect to backend. Refreshing when connection is ready.

mmelko:
Waiting screen in case of backend unavailability is added. Kaoto will try to fetch capabilities endpoint for 120s. 

close #826 

loading2

### DIFF
--- a/src/layout/AppLayout.tsx
+++ b/src/layout/AppLayout.tsx
@@ -1,14 +1,15 @@
 // @ts-ignore
 import logo from '../assets/images/logo-kaoto.png';
+import WaitingPage from '../components/WaitingPage';
 import { Page, PageHeader, SkipToContent } from '@patternfly/react-core';
-import { ReactNode } from 'react';
+import { ReactNode, useState } from 'react';
 import { useHistory } from 'react-router-dom';
 
 interface IAppLayout {
   children: ReactNode;
 }
-
 const AppLayout = ({ children }: IAppLayout) => {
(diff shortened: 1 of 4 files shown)


src/layout/AppLayout.tsx
```diff
@@ -35,7 +36,8 @@ const AppLayout = ({ children }: IAppLayout) => {
   );
   return (
     <Page mainContainerId={pageId} header={Header} skipToContent={PageSkipToContent}>
-      {children}
+      {!backendAvailable && <WaitingPage setBackendAvailable={setBackendAvailable} />}
+      {backendAvailable && children}
```
kahboom, line 40: Can we not just do `{backendAvailable ? children : <WaitingPage/>}`

mmelko: yeah we really should!

kahboom (approved):

Issue #586, opened 2022-07-12
Allow users to set the preferred way of Auto Layout


kahboom:
We should allow users to alternate between different Auto-Layouts (e.g. Horizontal, Vertical, freeform).

This would, in effect, override their existing Step Positions, so there should be a notification to the user to explain that.

~~On hold until #579 is resolved, as how the positions are stored will change.~~

Delawen:
Closed in #889

PR #889, opened 2022-11-03
feat(viz): layouting & autolayout

### LINKED ISSUE #871: [branches] Location of branches
### Describe the Bug

Branches steps are floating somewhere they shouldn't.

_No response_

1. Open Kaoto
2. In Settings, select `Camel Route`
3. [Optional] Add first step `timer`
4. [Optional] Add last step `log`
5. [Optional] Add intermediate step `choice`
6. In the source code editor, complete the route with this:
```
- from:
    uri: timer:null
    steps:
    - choice:
        when:
        - simple: '{{?foo}}'
          steps:
          - dynamic-router:
              expression:
                simple: ${body}
    - to:
        uri: log:null
```

### Screenshots or Videos

Firefox:
imagen

Chromium:
imagen

### Platform

- OS: Linux
- Browser: Chrome and Firefox
- Version: Latest `main`

kahboom:
## Changes

- distribute nodes automatically using dedicated layouting engine
- adds custom controls and ability to choose vertical/horizontal flow for visualization canvas
- centers the canvas initial node (again)
- temporarily disables dragging of nodes except for branch nodes
- **NOTE:** removes grouping of nodes for now to simplify testing, will be re-added in subsequent PR

implements #586, fixes #871 

### Horizontal Layout

Screen Shot 2022-11-03 at 11 02 34 am

### Vertical Layout

Screen Shot 2022-11-03 at 11 02 49 am

### Horizontal Layout with Branching

Screen Shot 2022-11-03 at 2 30 51 pm

### Vertical Layout with Branching
Screen Shot 2022-11-03 at 2 31 15 pm

### Custom Controls
The last two buttons are new controls for selecting the layout of choice.

Screen Shot 2022-11-03 at 11 03 09 am


package.json
```diff
@@ -44,6 +44,7 @@
     "@patternfly/react-table": "4.111.33",
     "@rhoas/app-services-ui-shared": "^0.16.4",
     "ajv": "^8.8.2",
+    "dagre": "^0.8.5",
```
apupier, line 47: If I'm not mistaken, it is this package which is used https://www.npmjs.com/package/dagre which is marked as deprecated https://github.com/dagrejs/dagre with last release which is 3 years old.

Do you feel confident that there would be a new release if needed?

Delawen:
I like how it looks

imagen

With a gibberish kamelet

```
apiVersion: camel.apache.org/v1alpha1
kind: Kamelet
metadata:
  annotations:
    camel.apache.org/kamelet.icon: whatever
  labels:
    camel.apache.org/kamelet.type: sink
  name: eip-action-sink
spec:
  definition:
    title: EIP Kamelet
    description: Used to test all EIP we implement
    properties: {}
  dependencies:
  - camel:core
  - camel:kamelet
  template:
    from:
      uri: kamelet:source
      steps:
      - loop:
          constant: '3'
          copy: true
          steps:
          - delay:
              expression:
                simple: ${body}
              async-delayed: true
      - choice:
          when:
          - simple: '{{?foo}}'
            steps:
            - dynamic-router:
                expression:
                  simple: ${body}
            - set-header:
                name: bar
                simple: foo
          - simple: '{{?bar}}'
            steps:
            - set-property:
                name: property
                simple: bar
            - remove-property:
                name: property
            - marshal:
                json:
                  library: Gson
          - simple: '{{?baz}}'
            steps:
            - transform:
                simple: baz
            - aggregate:
                correlation-expression:
                  simple: ${header.StockSymbol}
                aggregation-strategy: myAggregatorStrategy
                completion-size: 2
            - log:
                message: test
                logging-level: INFO
                log-name: yaml
          otherwise:
            steps:
            - set-body:
                simple: ola ke ase
            - remove-header:
                name: removeme
            - claim-check:
                operation: Push
                key: foo
                filter: header:(foo|bar)

```

src/components/Visualization.tsx
```diff
@@ -27,7 +29,11 @@ interface IVisualization {
 const Visualization = ({ toggleCatalog }: IVisualization) => {
   // `nodes` is an array of UI-specific objects that represent
   // the Integration.Steps model visually, while `edges` connect them
-  const defaultViewport: Viewport = { x: 10, y: 15, zoom: 1.2 };
+  const defaultViewport: Viewport = {
+    x: window.innerWidth / 2 - 80,
+    y: window.innerHeight / 2 - 160,
```
mmelko, line 34: why 80 and 160?

kahboom: That is the size of the node, which is fixed regardless of window size. And 160 is because it counts it from the center of the node, so you have to double it. I can change it to the use node width variable maybe so it's clearer.

package.json
kahboom: @apupier How about we merge what we have for now so that we can get feedback on branching and autolayout, and I can swap the layouting implementation underneath into next week? So as not to block this. Or do you think it'd be best to keep open until making this change?

apupier: Given that it is an highly awaited feature, I think we can go to merge with this dependency and create an issue with Critical(?) state to move to another implementation of layouting.

src/components/Visualization.tsx
mmelko: thanks for the answer.

PR #902, opened 2022-11-07
chore: bump kaoto version to 0.4.1

kahboom:
@apupier @Delawen - maybe we can do something about the fact that I always forget to bump the version in the `package.json` before a release. 😂  We need to do this in order to publish the correct npm package version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "kaoto-ui",
-  "version": "0.3.1",
+  "version": "0.4.1",
   "private": false,
   "federatedModuleName": "kaoto",
   "engines": {


apupier (approved):

Delawen:
We can have a script that does all the "generate new version" tasks. Can you write a list of actions to do (and commands if possible) so I generate it?

kahboom:
@Delawen sure, I think the main things are:
- Update the release notes, probably will have to be done manually if we want it to be formatted nicely?
- Update version in `package.json`
- Publish new version in npm
I can't think of anything else, but will let you know if I do.

PR #908, opened 2022-11-08
chore : package-ecosystem in dependabot for yarn shoudl use npm

apupier:
see https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates#supported-repositories-and-ecosystems

https://github.com/KaotoIO/kaoto-ui/network/updates
image


kahboom (approved):
thanks @apupier !

PR #916, opened 2022-11-08
chore: revert to previous scoped package name

kahboom:
@apupier will this be a problem for the vscode-plugin? this is so that we can publish to the kaoto org with the nice scoped name like we did previously: `@kaoto/kaoto-ui`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "kaoto-ui",
+  "name": "@kaoto/kaoto-ui",
   "version": "0.4.1",
   "private": false,
   "federatedModuleName": "kaoto",


apupier:
> will this be a problem for the vscode-plugin?

I do not know the implications. @thiagoelg changed this package name https://github.com/KaotoIO/kaoto-ui/commit/31f66fa3a5af9f7aeb663695e9beec5aab738619
So @thiagoelg , what are the implications?

kahboom:
I believe @thiagoelg only changed it to revert from `@thiagoelg/kaoto-ui` back to `kaoto-ui`. Probably we would just need to update the dependency name in the vscode repo to be `@kaoto/kaoto-ui`, unless there is something else I'm missing.

apupier (approved):

apupier:
created to align it in VS Code Kaoto project https://github.com/KaotoIO/vscode-kaoto/issues/66

kahboom:
thanks @apupier !

PR #925, opened 2022-11-10
fix(KaotoToolbar): Focus input when editing integration name. #768

### LINKED ISSUE #768: [UX] when editing the integration name the field is not automatically focused
**Describe the bug**
When clicking the edit button next to the integration name I have to click again the name text field to finally be able to edit the name.

**To Reproduce**
1. open Kaoto
2. Click on the edito button next to the integration name
3. try to type the new name without clicking the text field again

**Expected behavior**
Clicking the edit button should automatically put the focus on the text field and maybe even autoselect the current name so it can be easily overwritten by just starting to type a new name.

**Screenshots**
grafik

lordrip:
### Context
Currently, when clicking on the edit button labeled with a pencil icon, enables the edit mode but doesn't focus on the input field.

This commit adds a callback to be executed whenever the input field appears to set focus on it and avoid the need to click on it to start editing.

I've chosen a `ref={}` property combined with an `useCallback` hook to execute the focus only once per editing cycle. It seems that there's no other `InputField` at the same time competing with the focus, so probably this is not a hard requirement but still a good idea IMHO 🤔.

Relates to #768

### Before - No focus on editing
https://user-images.githubusercontent.com/16512618/201195461-a07fb4eb-7172-43dc-867f-378a406bff06.mp4

### After - Focusing Input field upon editing
https://user-images.githubusercontent.com/16512618/201195497-1accf208-3c0f-4272-9240-37d5f16a2815.mp4

### DIFF
--- a/src/components/KaotoToolbar.tsx
+++ b/src/components/KaotoToolbar.tsx
@@ -41,7 +41,7 @@ import {
   TimesIcon,
 } from '@patternfly/react-icons';
 import { useAlert } from '@rhoas/app-services-ui-shared';
-import { useEffect, useState } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 
 export interface IKaotoToolbar {
   toggleCatalog: () => void;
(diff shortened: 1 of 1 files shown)


lordrip:
As a side comment, not strictly relevant for this PR, but what are your thoughts on extracting this `InputField` into a dedicated component?

On the bright side, I think that it will ease testing this specific component while reducing the size of the `KaotoToolbar` component. Probably the downside would be having more pressing topics at hand 😅.

lhein:
Thanks for taking care, @lordrip !

kahboom:
>As a side comment, not strictly relevant for this PR, but what are your thoughts on extracting this InputField into a dedicated component?

@lordrip Yeah that probably could benefit from being moved to a separate component

kahboom (approved):
Looks good to me, thanks @lordrip !

PR #930, opened 2022-11-14
chore(ci): add job names to make status checks required

kahboom:
## Description
In order to list a status check as required, the job must contain a name, and the workflow name does not count.


Delawen (approved):

Delawen:
emojis :D

PR #932, opened 2022-11-14
chore: bump to v0.4.3, change to use dist

kahboom:
The PR aims to solve the issue with the NPM package not including all relevant build files, making it unconsumable by `vscode-kaoto`.

Changes:
~~First trying to change the `files` array to see if this will fix the issue of directories missing in `npm publish`, otherwise we will need to add an `npmignore` file.~~

- Bump `package.json` version to be able to republish the npm package
- Add step to `Publish NPM Package` GitHub action to run `yarn build:lib` so that the `lib` files get included
- Update `files` array in `package.json` to include the entire `dist` directory

### DIFF
--- a/.github/workflows/publish-npm-package.yml
+++ b/.github/workflows/publish-npm-package.yml
@@ -15,6 +15,7 @@ jobs:
           scope: '@kaoto'
       - run: yarn
       - run: yarn build
+      - run: yarn build:lib
       - name: 📦 Publish Package to npm
         run: yarn publish
         env:
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@kaoto/kaoto-ui",
-  "version": "0.4.2",
+  "version": "0.4.3",
   "private": false,
   "federatedModuleName": "kaoto",
   "engines": {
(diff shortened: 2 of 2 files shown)


apupier (changes requested):
I think it is still nto working, when doing:
`yarn clean`
`yarn build`
`npm pack`
the created tgz archive contains a dist folder but not the dist/lib folder

I have a `dist/lib` folder but no files in `dist` when doing:
`yarn clean`
`yarn build:lib`
`npm pack`

I have a `dist/lib` folder and files in `dist` when doing:
`yarn clean`
`yarn build`
`yarn build:lib`
`npm pack`


In fact, the `yarn build`, is generating the direct children files in dist folder. And `yarn build:lib` is generating the content of the `dist/lib` folder. No need to go to `npm pack` to notice it

kahboom:
@apupier okay will test it again after my appointment

kahboom:
It seems this PR is actually addressing the problem and we won't need to try with an `.npmignore` file. I tried the steps mentioned by @apupier , both with the `main` branch and this one. If we kept the `package.json` as `files: ["dist/lib"]` then the main `dist` files do not get included in the package.

I've also added a commit to run `yarn build:lib` as an additional step in the `Publish NPM Package` GH action so that the library files are included.

apupier (approved):

PR #936, opened 2022-11-15
feature(standalone): Add a job that creates jar and native standalone for releases

Delawen:
What this does is that when a release is published, it generates native executables and a jar that are standalones. Meaning, they contain both backend (latest version that matches major and minor) and frontend (the one from the release).

### DIFF
--- a/.github/workflows/add-standalone-jar-to-release.yml
+++ b/.github/workflows/add-standalone-jar-to-release.yml
@@ -1,0 +1,131 @@
+name: 👩‍💻 Add Standalone Jar to Release
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  build-frontend:
+    name: 👷‍♀️ Build Frontend
+    runs-on: ubuntu-latest
+    steps:
+      - name: 🚟 Checkout backend
+        uses: actions/checkout@v3
+      - name: 🚀 Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
+          registry-url: 'https://registry.npmjs.org'
+          scope: '@kaoto'
+      - name: 🛸 Build Frontend
+        run: yarn install --mode=skip-build && KAOTO_API="" yarn run build
+      - name: 📩 Save dist folder
+        uses: actions/upload-artifact@v3
+        with:
+          name: dist
+          path: /app/kaoto-ui/dist/
+          retention-days: 1
+
+  build-standalone-jar:
+    name: 👩‍🏭 Build standalone jar
+    runs-on: ubuntu-latest
+    steps:
+      - name: 🛂 Find the right frontend major and minor version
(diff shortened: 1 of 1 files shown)


.github/workflows/add-standalone-jar-to-release.yml
```diff
@@ -0,0 +1,132 @@
...
+          path: ~/.m2
+          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ runner.os }}-m2
+      - name: 🤳 Build Jar
+        run: mvn install -DskipTests
+      - name: 📦 Upload binaries to release
+        uses: svenstaro/upload-release-action@v2
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file: api/target/api-1.0.0-SNAPSHOT.jar
```
Delawen, line 70: I know this version is hardcoded, I will fix first on the backend to have the proper version in the name and then will drag the version variable here too.

kahboom (approved):

.github/workflows/add-standalone-jar-to-release.yml
```diff
@@ -0,0 +1,131 @@
...
+      - name: 👩🏻‍💼 Move frontend to proper folder
+        run: 'mv api/src/main/resources/META-INF/dist/ api/src/main/resources/META-INF/resources/'
+      - name: 🥸 Set up JDK
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'adopt'
+          java-version: 17
+          check-latest: true
+      - name: 🔥 Cache Maven packages
+        uses: actions/cache@v1
```
apupier, line 59: there is a v3

.github/workflows/add-standalone-jar-to-release.yml
```diff
@@ -0,0 +1,131 @@
...
+            file: api-1.0.0-SNAPSHOT-runner
+          - os: windows-latest
+            asset_name: kaoto-windows-amd64
+            file: api-1.0.0-SNAPSHOT-runner.exe
+          - os: macos-latest
+            asset_name: kaoto-macos-amd64
+            file: api-1.0.0-SNAPSHOT-runner
+    steps:
+      - name: 🛂 Find the right frontend major and minor version
+        run: 'echo "frontend_version=echo "${{ env.release.tag_name }}" | sed -En 's/v([0-9]+\.)([0-9]+\.)?([0-9]+)/v\1\2/p'" >> $GITHUB_ENV'
```
apupier, line 91: I have this validation error:

image

apupier: there are simple quotes used at places, I think the one around parameter of sed shoudl be escaped

apupier: maybe need to escape with another single-quote (so we have two simple-quote):

`'echo "frontend_version=echo "${{ env.release.tag_name }}" | sed -En ''s/v([0-9]+\.)([0-9]+\.)?([0-9]+)/v\1\2/p''" >> $GITHUB_ENV'`

based on https://yaml.org/spec/1.2.2/#single-quoted-style

PR #972, opened 2022-11-25
feat(codeEditor): Disabling autosync

mmelko:
This PR introduces three modes how code editor will work. This currently cannot be changed in the settings.. it can be a following PR if we will to have it configurable.
- `FREE_EDIT`:
  - Editor is Read only, and new Edit button is added into controls:
     Screenshot 2022-11-25 at 10 35 21
  - After clicking on edit, a modal window will appear when user can edit the code and sync changes:
     Screenshot 2022-11-25 at 10 35 03
- `READ_ONLY`: 
  - Editor is read_only and edit modal cannot be displayed. This might be useful in Vscode plugin, or when we don't want to change the code at all.
     Screenshot 2022-11-25 at 10 33 47
-  `TWO_WAY_SYNC`:
  -  Old behaviour when editor is editable, and code is automatically replaced if it's invalid:
      Screenshot 2022-11-25 at 11 00 43

### DIFF
--- a/src/components/SourceCodeEditor.tsx
+++ b/src/components/SourceCodeEditor.tsx
@@ -75,6 +75,7 @@ const SourceCodeEditor = (props: ISourceCodeEditor) => {
       );
     });
 
+    editor?.revealLine(editor.getModel()?.getLineCount() ?? 0);
     editorRef.current = editor;
   };
 
@@ -172,6 +173,7 @@ const SourceCodeEditor = (props: ISourceCodeEditor) => {
         code={sourceCode ?? props.initialData}
         className="code-editor"
         height="80vh"
+        width={'100%'}
         onCodeChange={debounced}
         language={(props.language as Language) ?? Language.yaml}
         onEditorDidMount={handleEditorDidMount}
@@ -181,9 +183,14 @@ const SourceCodeEditor = (props: ISourceCodeEditor) => {
         isDarkTheme
         isDownloadEnabled
         isLanguageLabelVisible
+        allowFullScreen={true}
         isUploadEnabled
         options={{
           readOnly: props.mode === CodeEditorMode.READ_ONLY || !props.editable,
+          scrollbar: {
+            horizontal: 'visible',
+            vertical: 'visible',
+          },
         }}
       />
     </StepErrorBoundary>
--- a/src/components/SourceCodeEditor.tsx
+++ b/src/components/SourceCodeEditor.tsx
@@ -1,11 +1,17 @@
 import { fetchIntegrationJson, fetchIntegrationSourceCode } from '@kaoto/api';
 import { StepErrorBoundary } from '@kaoto/components';
 import { useIntegrationJsonStore, useIntegrationSourceStore, useSettingsStore } from '@kaoto/store';
-import { IIntegration } from '@kaoto/types';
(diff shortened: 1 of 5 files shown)


Delawen:
I love being able to keep the double sync still :)

Advanced function

mmelko:
Yeah, I still believe we will get back to it when we improve our validation.

src/components/SourceCodeEditor.test.tsx
```diff
@@ -5,7 +5,7 @@ import { render } from '@testing-library/react';
 
 describe('SourceCodeEditor.tsx', () => {
   test('component renders correctly', () => {
-    render(<SourceCodeEditor language={Language.yaml} />);
+    render(<SourceCodeEditor language={Language.yaml} editAction={() => {}} />);
```
kahboom, line 8: You should just be able to do `jest.fn()` here to mock a function

src/components/SourceCodeEditorModal.tsx
```diff
@@ -0,0 +1,32 @@
...
+        <Alert
+          title="Warning"
+          variant="warning"
+          actionClose={<AlertActionCloseButton onClose={() => setAlertVisible(false)} />}
+        >
+          Any invalid code will be replaced after sync. If you don't want to lose your changes
+          please make a backup.
+        </Alert>
+      )}
+      <SourceCodeEditor editAction={() => {}} editable={true} mode={CodeEditorMode.FREE_EDIT} />
```
kahboom, line 29: Why do we have the `editAction` prop if it's not doing anything?

kahboom:
Some additional minor feedback here:

Screen Shot 2022-11-28 at 11 23 55 am

1- Can we say "Cannot edit in read-only mode" instead of "read-only editor"? Just so people know there is more than one mode, in case it wasn't clear enough.


Screen Shot 2022-11-28 at 11 32 11 am
2- Can we highlight the Edit button in green, or maybe blue (some color to make it stand out), like we do for save? Green might not be the right color, but basically I just want people to "know" where to put it into Edit mode.

mmelko:
1. It should be possible. That is actually an default message from Monaco. Do we want to have modes adjustable by user? If yes that make sense. Currently they aren't
2. Good idea:)

src/components/SourceCodeEditorModal.tsx
mmelko: So, now it isn't doing anything.. but when we want to show diff for example, I can imagine I'd do it there. But yeah it doesn't make sense to have it know. I will add some //TODOs not to forget if you agree.

mmelko:
@kahboom I updated the PR please check

src/components/SourceCodeEditor.tsx
```diff
@@ -57,24 +67,41 @@ const SourceCodeEditor = (props: ISourceCodeEditor) => {
   };
 
   const handleEditorDidMount = (editor: EditorDidMount['editor']) => {
+    const messageContribution: any = editor?.getContribution('editor.contrib.messageController');
```
kahboom, line 70: Quick question-- why `editor?` instead of just `editor`, considering it's a required parameter?

mmelko: thanks you're right

kahboom (approved):

PR #1003, opened 2022-12-08
fix(UI/UX/Accessibility): improving the coloring for a better visibility

### LINKED ISSUE #985: [UI/UX/Accessibility] Buttons on canvas figures have a too light coloring
### Please describe the task that needs to be done

Having the contextual buttons like "Delete" or "+" white on grey makes them really hard to see and recognize. Can we improve the coloring for a better visibility? 

grafik

miram13:
Which issue(s) this PR fixes: #985 

For easier accessibility, the colouring is done in this manner.

Screenshot from 2022-12-06 11-56-35

### DIFF
--- a/src/components/PlusButtonEdge.css
+++ b/src/components/PlusButtonEdge.css
@@ -1,8 +1,20 @@
-.trashButton, .plusButton {
+.plusButton {
     width: 20px;
     height: 20px;
-    background: var(--pf-global--Color--light-300);
-    color: white;
+    background: var(--pf-global--Color--light-200);
+    color: rgb(38, 134, 75);
+    border: 1px solid #fff;
+    cursor: pointer;
+    border-radius: 50%;
+    font-size: 12px;
+    line-height: 1;
+    padding: 1px;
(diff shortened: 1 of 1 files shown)


Delawen:
This is a suggestion coming from a backender, so don't feel forced to hear me. 

Can we have less bright colors and then make them brighter when the mouse is over it?

Also, the trash icon in red will be difficult to see for visually impaired people. Maybe we can look for a different icon (not an `x` because that may be a problem with dyslexia and the `+` button)? Do we have a different 🗑️ icon? Or maybe a `-` to contrast the `+`?

kahboom:
@Delawen - we meet with UXD again next week, right? I think we should leave it for then.

I agree with you on making it less bright and brighter when you hover over it. I think @lhein wanted it to be more visible, but I don't think these buttons should be so visible for every single step unless you are hovering or specifically looking for them. Otherwise it will look clunky.

Not sure about the bag icon, since it's more for shopping carts, but UXD will know more I think.

PR #1010, opened 2022-12-10
fix(Visualization): Initial node isn't always centered

### LINKED ISSUE #952: Visualization: Initial node isn't always centered
### Describe the Bug

The initial "ADD A STEP" node isn't centered. The React Flow library maintainer suggests the following for centering the node:

>when you want to place a node in the center you need to use the project function to convert the screen coordinates to the internal coordinate system.

1. Measure the size of the renderer via `getBoundingClientRect` for example ([link](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect))
2. Put the center of the dimensions in the [project function](https://reactflow.dev/docs/api/hooks/). For a `1200x600px` renderer: `const xy = project({ x: 600, y: 300 })`
3. Use the `xy` position for your first node

A bonus would be to also re-center it when switching between horizontal/vertical flow.

1. Load Kaoto
2. See that initial node may appear centered but isn't

### Platform

- OS: macOS
- Browser: Chrome
- Version: 0.5.0-dev

javierlopezdeancos:
This PR fix and Close: #952 

## Summary of changes
- [x] Add a new `useEffect` to detect if nodes only have the initial "ADD A STEP" node to set their position in the React Flow viewport

If user switch between horizontal/vertical flow, the position should be re calculated in the `useEffect`

## Screenshots
kaoto-center-1
kaoto-center-2

### DIFF
--- a/src/components/Visualization.tsx
+++ b/src/components/Visualization.tsx
@@ -15,6 +15,7 @@ import {
   findStepIdxWithUUID,
   getLayoutedElements,
 } from '@kaoto/services';
+import { containsAddStepPlaceholder } from '@kaoto/services';
 import { useIntegrationJsonStore, useVisualizationStore } from '@kaoto/store';
 import { IStepProps, IViewData, IVizStepPropsEdge, IVizStepNode } from '@kaoto/types';
 import { useEffect, useMemo, useRef, useState } from 'react';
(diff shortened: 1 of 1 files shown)


Delawen:
@javierlopezdeancos can you check both warnings from sonarcloud?

Delawen:
I think we need @kahboom or @mmelko to take a look at this, this is beyond what I fear to touch the frontend at this point :)

Thank you @javierlopezdeancos !

javierlopezdeancos:
Don't worry @Delawen, I'll wait for your suggestions delighted!

kahboom (approved):

kahboom:
thank you @javierlopezdeancos for your contribution, this looks good to me :)

javierlopezdeancos:
A pleasure to contribute, I will try to do it regularly

PR #1019, opened 2022-12-13
fix(release): Fixing creation of release executables and add manual trigger

Delawen:
So if the release executable creation fails, we can re-launch it.

### DIFF
--- a/.github/workflows/add-standalone-jar-to-release.yml
+++ b/.github/workflows/add-standalone-jar-to-release.yml
@@ -1,6 +1,12 @@
-name: 👩‍💻 Add Standalone Jar to Release
+name: 👩‍💻 Add Builds to Release
 
 on:
+  workflow_dispatch:
+    inputs:
+      tag_name:
+        description: 'Tag to use (frontend)'
+        required: true
+        type: string
   release:
     types: [published]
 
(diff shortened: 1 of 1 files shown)


Delawen:
This is one of these "this could have never worked before, what were I thinking?" situations.

apupier (changes requested):
it doesn't seem to be working, see https://github.com/apupier/kaoto-ui/actions/runs/3685775351/jobs/6237194824

Also even with the first step which is green, in fact there is a warning message that mention that no file s were found in the dist folder https://github.com/KaotoIO/kaoto-ui/actions/runs/3685129434/jobs/6235707558#step:5:11 Given that the artifact is tried to be used in a later stage, it won't be able to be working

Delawen:
Damn, in my local it works! Thanks

Delawen:
Getting closer but not yet https://github.com/Delawen/kaoto-ui/actions/runs/3687248060

Delawen:
The Windows build is not working because the script to find the backend version don't work on Windows: https://github.com/Delawen/kaoto-ui/actions/runs/3692831943/jobs/6252153288

Not sure how to work around that.

apupier:
For Windows, it will be something like:

https://github.com/KaotoIO/kaoto-backend/blob/42a4e102138441087f281b5e31c0fbb47097e9cd/.github/workflows/publish-release.yml#L46-L49

Delawen:
The missing parts on Windows are a `sed`

https://github.com/Delawen/kaoto-ui/actions/runs/3692913265

and a `git` command

https://github.com/Delawen/kaoto-ui/actions/runs/3692913265

I think the git command should be fine, but the sed...

apupier (dismissed):

Issue #958, opened 2022-11-22
Visualization: Camel routes aren't being rendered


kahboom:
### Describe the Bug

Camel routes are not being show in the Visualization, even if you select the correct DSL beforehand. There is also an error in the browser console:

```
Cannot read properties of null (reading 'name')
    at eval (SourceCodeEditor.tsx?11c6:38:1)
```

### Steps to Reproduce the Bug or Issue

1. Go to Settings
2. Change DSL to Camel Route
3. Click save
4. Open the Source Code Editor
5. Paste a Camel route YAML
6. See it doesn't render the route in the Visualization

You can try with one of these YAMLs:

```yaml
- from:
    uri: "timer:start"
    steps:
    - choice:
        precondition: true
        when:
        - simple: "{{?red}}"
          steps:
          - to: "log:red"
        - simple: "{{?blue}}"
          steps:
          - to: "log:blue"
        otherwise:
          steps:
          - to: "log:info"
```

```yaml
- from:
    uri: "direct:start"
    steps:
    - set-property:
        name: test
        expression:
          simple: "${body}"
    - to: "mock:result"
```

```yaml
- from:
    uri: "direct:route"
    steps:
    - filter:
        simple: "${body.startsWith(\\"a\\")}"
        steps:
          - to: "mock:filter"
    - to: "mock:route"
```

### Screenshots or Videos

![Screen Shot 2022-11-22 at 12 38 27 pm](https://user-images.githubusercontent.com/3844502/203316039-7262c9fe-17b2-4ac0-8f8e-2ea9bd3efec3.png)

### Platform

- OS: macOS
- Browser: Chrome
- Version: 0.5.0-dev

lordrip:
Hi @kahboom, this problem seems to be caused because [this check](https://github.com/KaotoIO/kaoto-ui/blob/47d5c7aa6022f29ced8878f855c4b1d0038e92a8/src/components/SourceCodeEditor.tsx#L56-L59)
```typescript
          if (res.metadata.name != '') {
            settings.name = res.metadata.name;
            setSettings({ name: res.metadata.name });
          }
```

using the `YAML` files that you provided, the `metadata` field is `null`, which is different than `<empty-string>`, therefore :boom:, but this is the easy part, I'm preparing a pull request to update it.

On the other hand, it seems that this fix is ~~uncovering~~ causing another bug (maybe a hidden feature), which is seeing the steps rendered on the screen but keeping the code editor empty.

[...]

Issue #973, opened 2022-11-25
Code editor isn't rendering in the test.


mmelko:
### Please describe the task that needs to be done

It appears that current code editor isn't really rendered. This might be actually an issue with Monaco Jest or missing some dependencies or mocks. 

This is current render of SourceCodeEditor component in the test:
```html
body>
  <div>
    <textarea />
  </div>
</body>
```

Delawen:
@lordrip is this still happening? How can I check?

lordrip:
Yes, it's still happening. This is because the `monaco` editor is a bit complex to run it through the JSDom environment.

We're [mocking the whole component here](https://github.com/KaotoIO/kaoto-ui/blob/d30e611a509a08d2268deb63ae7cf90b0ef282d3/src/__mocks__/reactCodeEditorMock.js)

In order to fix this, the idea is to split the buttons from the code editor itself, so we can at least check the interaction between Kaoto and the code editor.

PR #1041, opened 2022-12-21
fix: Camel routes not being rendered

lordrip:
Currently, Camel routes don't have a name property, causing an exception whenever the source code editor attempts to access it to propagate the name to the UI.

This commit adds an extra validation to ensure that the metadata field is not `null` and that the name is a non-empty string.

handles [#958](https://github.com/KaotoIO/kaoto-ui/issues/958)

### DIFF
--- a/src/components/SourceCodeEditor.tsx
+++ b/src/components/SourceCodeEditor.tsx
@@ -53,7 +53,7 @@ const SourceCodeEditor = (props: ISourceCodeEditor) => {
       fetchIntegrationJson(incomingData, settings.dsl.name)
         .then((res: IIntegration) => {
           let tmpInt = res;
-          if (res.metadata.name != '') {
+          if (typeof res.metadata?.name === 'string' && res.metadata.name !== '') {
             settings.name = res.metadata.name;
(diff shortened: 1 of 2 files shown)


lordrip:
I've tried to make a test to replicate the existing issue but I couldn't give that the `CodeEditor` is being mocked as a `<textarea>` tag. In the meantime, I've created this pull request as a mean to discuss what's happening with issue #958  

I can think about two alternatives:

1. Split `CodeEditor` as `CodeEditorToolbar` and `CodeEditor`, so the toolbar is kept to make some assertions about the behavior of Saving / Updating the code :thinking: 

2. Keep the `CodeEditor` as it is but enhance the mock to include raw buttons to fire some events that allow generating desired behaviors

I'm open to suggestions :smiley:

mmelko (approved):

mmelko:
@lordrip this looks good thanks! 

To your question I believe we should mock full code editor at some point to being able to test other functionality too. We can have some discussion here: https://github.com/KaotoIO/kaoto-ui/issues/973 . (tittle of the issue needs to change)

PR #1119, opened 2023-01-17
enhancement: remove kaoto banner and move logo to the toolbar

### LINKED ISSUE #937: Get rid of Kaoto banner and move logo to the toolbar

Currently the toolbar occupies quite lot of space that isn't really useful. I believe we should get rid of it and move the logo somewhere to the toolbar.  This will also look better in the vscode. 
Current: 
Screenshot 2022-11-15 at 16 43 53

mmelko:
close #937
I moved the logo, changed the font and styles of few components to accommodate new height. Also link to `/` from the logo was removed because we don't have any pages so I don't think it's necessary. 

UI after the change:
Screenshot 2023-01-17 at 15 18 40


apupier (commented):
Codewise, I'm not familiar enough with neither the codebase, React and CSS to review. But I really like what I see in the screenshot

kahboom (changes requested):
I think the measurements need to be adjusted because otherwise it would introduce a regression, or a bug we previously fixed where if you scroll too far then you can't scroll back up:

Kapture 2023-01-18 at 14 41 18

mmelko:
@kahboom please check again.

kahboom (approved):
It's perfect, thanks @mmelko !

Issue #1133, opened 2023-01-19
Visualization: Show when steps are part of a branch


kahboom:
## Problem
Currently, we are showing branches and their steps in exactly the same way as a completely linear integration, which isn't normally such a problem if the step has more than one branch. However, if a step only has one branch, it's shown inline, like the steps to the right of `filter`, below:

![Screen Shot 2023-01-19 at 11 41 54 am copy](https://user-images.githubusercontent.com/3844502/213450847-60507507-bbe2-46b8-8197-7c7d22954feb.png)

It's not possible to know, unless you look at the data structure, that those steps are actually inside of a branch that belong to `filter`.

![Screen Shot 2023-01-19 at 11 41 54 am copy 3](https://user-images.githubusercontent.com/3844502/213451153-dc5e9fea-f855-485f-9878-6b3efde2deeb.png)

Data structure shortened for brevity:

```json
{
      "name": "filter",
      "branches": [
        {
          "steps": [
            {
              "name": "set-body",
            },
            {
              "name": "unmarshal",
            },
            {
              "name": "circuit-breaker",
              "branches": [
                {
                  "steps": [
                    {
                      "name": "enrich",
                    },
                    {
                      "name": "dropbox",
                    },
                    {
                      "name": "convert-body-to",
                    }
                  ],
                },
                {
                  "steps": [
                    {
                      "name": "log"
                    }
                  ]
                }
              ],
            }
          ],
        }
      ]
    },
```

So if you were to replace the `filter` step with `atlasmap` (for example), the integration would now look like this:

![Screen Shot 2023-01-19 at 11 42 40 am copy 2](https://user-images.githubusercontent.com/3844502/213452391-16d55ae2-de8d-45b5-8bf2-76601d86563b.png)

The user would likely be confused as to why so many steps were deleted, unless they knew the data structure or really understood the flow of the integration.

In general, though, it would improve the user experience to show a better visual representation of how steps are actually related.

## Proposed Solution
There are several options, all of which should be discussed with the UXD team. These are a few of them:

1. Highlight all edges of a step and its child steps on hover (i.e. mousing over `filter` or any of the child branch steps, like `dropbox` would shower a thicker, maybe colored edge through all of them, or animate the edge)
2. Bring back the branch containers that we previously had, which would put all the child steps inside of a group
3. We can also move branches to extend vertically, rather than horizontally (or vice versa, depending on the auto-layout choice), but this would open up the canvas to several consequences down the line, such as node collisions

Issue #786, opened 2022-10-10
Branching: Build branches visually


kahboom:
## Description
At the moment we can only render branches, but we can't build them ourselves from the low code editor. We should be able to, on adding an EIP to the integration, configure the branches and steps for it.

This task will require a lot of planning, but likely requires #586 beforehand (TBD), as well as other branch issues.

Requires #1158 to be completed first.

Epic: #140

PR #1140, opened 2023-01-20
feat(viz): add ability to insert branch steps

### LINKED ISSUE #1114: Add ability to insert a step between two branch steps
Epic: https://github.com/KaotoIO/kaoto-ui/issues/783

kahboom:
This PR adds the ability to insert a step in between two branch steps. Resolves #1114.

## Changes
- Add check in mini catalog to validate inserting branch steps
- Add handling in mini catalog step selection for branch steps
- Enable `insert` edge type for branch steps
- Extract `insertStep` logic into its own testable function
- Improve query params for API request to be more accurate with source/target step
- Improve typings

## Screenshots

Kapture 2023-01-20 at 11 09 00-2

### DIFF
--- a/src/components/PlusButtonEdge.tsx
+++ b/src/components/PlusButtonEdge.tsx
@@ -1,14 +1,15 @@
 import './PlusButtonEdge.css';
 import { MiniCatalog } from '@kaoto/components';
-import { findStepIdxWithUUID, insertableStepTypes } from '@kaoto/services';
-import { useIntegrationJsonStore } from '@kaoto/store';
-import { IStepProps } from '@kaoto/types';
+import { findStepIdxWithUUID, insertableStepTypes, insertStep } from '@kaoto/services';
+import { useIntegrationJsonStore, useNestedStepsStore } from '@kaoto/store';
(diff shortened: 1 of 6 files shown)


mmelko (changes requested):
If only one step is left in the branch, I'm not able to add any step into that branch anymore: see: Screenshot 2023-01-20 at 13 37 27

kahboom:
@mmelko but how would you know whether the user wants to add a step to the branch or to the main integration? for this thing I think we would need to have a design review and also be working on building branches visually #786 .

mmelko:
So the simplest would be to have 2x `+` one in the branch and second at the integration level before the first step of the first step after branches (filter step in this cas)

kahboom:
I don't see a filter step, but I think I know what you mean. But that's why I mentioned prepend step would need to be supported, and right now it's not. To me that should be in a separate PR.

Actually I'm thinking about it and even if we allowed appending/prepending steps in a branch, it will look really weird to users until we solve #1133 , because it will have two + buttons in between two steps, the user won't know the difference between those buttons, or know that one is to add inside of a branch and the other for outside of the branch.

mmelko (approved):

PR #1165, opened 2023-01-25
fix(CodeEditor): fix missing actions after cleanup the source code

### LINKED ISSUE #1132: Revert code editor to previous state

We first need to fix the issue with data loss, which will require a discussion with UXD, but the current code editor is making for quite a bad user experience.

Previously, we used to be able to:
1) write the code, and it would render immediately

Now, we the process looks like this:
1) click the edit button (once you know where it is)
2) write the code
3) press the save button (once you know where it is)
4) close the success notification to be able to then
5) press the close button on the modal

That's an additional 4-5 steps that we didn't have to do before.

Ideally, we will fix the issue with data loss and revert to the previous state of the code editor, maybe with some additional handling of validation.

### LINKED ISSUE #1147: Clearing the code editor removes save & clear buttons
### Describe the Bug

If you press the clear button on the code editor (in edit mode) and then make changes, the save button disappears completely and you can't save at all. The same happens for the clear button.

_No response_

1. Go to the code editor
2. Click edit mode
3. Press the clear button
4. Press "Start from scratch"
5. Make changes
6. The save button doesn't reappear

Repeat steps for the clear button.

### Screenshots or Videos

Code editor clear

### Platform

- OS: macOS
- Browser: Chrome
- Version: `0.6.1-dev`

mmelko:
disable CodeEditorModal and enable edit mode in a left panel again 
add default value for validationSchema in the settings
 hide left panel when settings are opened
fix #1132 
fix #1147

### DIFF
--- a/src/components/KaotoToolbar.tsx
+++ b/src/components/KaotoToolbar.tsx
@@ -48,12 +48,13 @@ import { useCallback, useEffect, useState } from 'react';
 export interface IKaotoToolbar {
   toggleCatalog: () => void;
   toggleCodeEditor: () => void;
+  hideLeftPanel: () => void;
 }
 function LogoImg() {
   return <img data-testid={'kaoto-logo'} src={logo} alt="Kaoto Logo" style={{ height: '30px' }} />;
 }
 
-export const KaotoToolbar = ({ toggleCatalog, toggleCodeEditor }: IKaotoToolbar) => {
+export const KaotoToolbar = ({ toggleCatalog, toggleCodeEditor, hideLeftPanel }: IKaotoToolbar) => {
   const { settings, setSettings } = useSettingsStore((state) => state);
   const { sourceCode, setSourceCode } = useIntegrationSourceStore((state) => state);
   const deleteIntegration = useIntegrationJsonStore((state) => state.deleteIntegration);
@@ -169,7 +170,10 @@ export const KaotoToolbar = ({ toggleCatalog, toggleCodeEditor }: IKaotoToolbar)
     <DropdownItem
       key="settings"
       data-testid={'kaotoToolbar-kebab__settings'}
-      onClick={() => setExpanded({ ...expanded, settingsModal: !expanded.settingsModal })}
+      onClick={() => {
+        setExpanded({ ...expanded, settingsModal: !expanded.settingsModal });
+        hideLeftPanel();
+      }}
     >
       Settings
     </DropdownItem>,
--- a/src/components/SourceCodeEditor.tsx
+++ b/src/components/SourceCodeEditor.tsx
@@ -4,14 +4,9 @@ import { useIntegrationJsonStore, useIntegrationSourceStore, useSettingsStore } 
 import { CodeEditorMode, IIntegration } from '@kaoto/types';
 import { usePrevious } from '@kaoto/utils';
 import { CodeEditor, CodeEditorControl, Language } from '@patternfly/react-code-editor';
-import {
-  CheckCircleIcon,
-  EraserIcon,
-  PencilAltIcon,
-  RedoIcon,
-  UndoIcon,
-} from '@patternfly/react-icons';
-import { ReactNode, useEffect, useRef } from 'react';
+import { Alert } from '@patternfly/react-core';
+import { CheckCircleIcon, EraserIcon, RedoIcon, UndoIcon } from '@patternfly/react-icons';
+import { useEffect, useRef, useState } from 'react';
 import EditorDidMount from 'react-monaco-editor';
 import { useDebouncedCallback } from 'use-debounce';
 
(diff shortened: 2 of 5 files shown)


src/components/KaotoToolbar.tsx
```diff
@@ -169,7 +170,10 @@ export const KaotoToolbar = ({ toggleCatalog, toggleCodeEditor }: IKaotoToolbar)
     <DropdownItem
       key="settings"
       data-testid={'kaotoToolbar-kebab__settings'}
-      onClick={() => setExpanded({ ...expanded, settingsModal: !expanded.settingsModal })}
+      onClick={() => {
```
kahboom, line 173: Did you forget to remove this extra space? Just this is causing the code coverage to go down, which is so funny. 😂

src/components/SourceCodeEditor.tsx
```diff
@@ -20,8 +14,9 @@ interface ISourceCodeEditor {
   language?: Language;
   theme?: string;
   mode?: CodeEditorMode | CodeEditorMode.FREE_EDIT;
+  schemaUri?: string;
   editable?: boolean | false;
-  editAction?: (code: string, event?: any) => void;
+  //01editAction?: (code: string, event?: any) => void;+-\
```
kahboom, line 19: Did you forget to remove this?

src/components/SourceCodeEditor.tsx
```diff
@@ -99,12 +92,16 @@ const SourceCodeEditor = (props: ISourceCodeEditor) => {
   };
 
   const debounced = useDebouncedCallback((value: string) => {
-    // if editor is in read only mode file can be still uploaded.
-    if (props.mode !== CodeEditorMode.FREE_EDIT) {
+    if (props.mode == CodeEditorMode.TWO_WAY_SYNC) {
```
kahboom, line 95: Can we not use `===` here?

kahboom (changes requested):
I've requested a few changes, but I also wanted to ask if we can add in a note to the user when the code editor is dirty, to ask them to save their changes and warn them about the loss of invalid data?

That way like with the old approach they know something is going to change in their code, and also that they need to save it. But it should only show once the editor is dirty.

src/components/SourceCodeEditor.tsx
mmelko: did you mean `==` ? because that's what I'm using. If yes what should I use instead ?

apupier: usually in TypeScript and JavaScript world, we prefer to use the triple equal `===` instead of the double equal `==`. the double equal can do some weird type transformation.

apupier: https://rules.sonarsource.com/typescript/RSPEC-1440

mmelko:
@kahboom  this will appear only if the code is different from the synced one and disappear after the sync or reverting the code.

Screenshot 2023-01-27 at 10 09 31

mmelko:
@kahboom  this will appear only if the code is different from the synced one and disappear after the sync or reverting the code.

Screenshot 2023-01-27 at 10 09 31

src/components/SourceCodeEditor.tsx
```diff
@@ -49,33 +48,29 @@ const SourceCodeEditor = (props: ISourceCodeEditor) => {
...
-    }
+    // update integration JSON state with changes
+    fetchIntegrationJson(incomingData, settings.dsl.name)
+      .then((res: IIntegration) => {
+        let tmpInt = res;
+        if (typeof res.metadata?.name === 'string' && res.metadata.name !== '') {
+          settings.name = res.metadata.name;
+          setSettings({ name: res.metadata.name });
+        }
+        tmpInt.metadata = { ...res.metadata, ...settings };
```
kahboom, line 59: Eventually we're gonna have to stop mutating so much data to prevent bugs :( this is my fault though for starting this lol

src/components/SourceCodeEditor.tsx
kahboom: Yep it's what @apupier said

src/components/SourceCodeEditor.tsx
```diff
@@ -175,37 +164,37 @@ const SourceCodeEditor = (props: ISourceCodeEditor) => {
...
-  } else if (props.mode === CodeEditorMode.FREE_EDIT && props.editable) {
-    customControls = [UndoButton, RedoButton, ClearButton, UpdateButton];
-  }
-
   return (
     <StepErrorBoundary>
+      {syncedCode !== sourceCode && (
+        <Alert
+          title="Any invalid code will be replaced after sync. If you don't want to lose your changes
+          please make a backup."
```
kahboom, line 176: What I was looking for was more a notice to let users know that they need to press the save button, not necessarily make a backup. They might not know it or even know where it is. I kind of wish we had a normal save button so it would be more obvious, but even a message would help.

kahboom (approved):

PR #1168, opened 2023-01-25
chore(ci): add codecov.yml

kahboom:
Add a `codecov.yml` file with configuration settings, and makes Codecov informational, rather than gating pull requests and failing the status: https://docs.codecov.com/docs/commit-status#informational

note: I'm actually on the fence about this, as I don't want us to let our code coverage slip either, but sometimes the coverage change is -0.1% and it still fails. I know we can change the threshold, but I figured maybe making it informational makes more sense? Let me know what you think.

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -1,0 +1,5 @@
+coverage:
+  status:
+    project:
+      default:
+        informational: true


apupier:
Is the gating is only on the red/green status but we can merge anyway?

I think it is a good thing to have the red mark even for -0.1%. At least, it is forcing to think about it seriously and provide a justification why we evaluate that there is no need to test (some parts of) the added code.

kahboom:
@apupier thanks, in that case do you think increasing the threshold to 0.50% would make sense, or even then we shouldn't have a failed status?

apupier:
> @apupier thanks, in that case do you think increasing the threshold to 0.25% or 0.50% would make sense, or even then we shouldn't have a failed status?

I think we should always have the failed status as soon as the coverage is decreasing. To force either to write the test or to explain why it is not worthy/not possible.

PR #1186, opened 2023-01-27
Add search and filter for Resources List

### LINKED ISSUE #587: Search & Filter for Resources List
Currently, we allow users to sort by Name and Date/Time. Users should be able to sort (also by Type), search, and filter from the Deployment/Resources list.

tplevko:
Closes https://github.com/KaotoIO/kaoto-ui/issues/587

Changed the deployments modal size to be static, so it doesn't shrink in case there are few elements in the list.

Screenshot from 2023-01-27 17-28-47
Screenshot from 2023-01-27 17-29-07

### DIFF
--- a/src/components/DeploymentsModal.tsx
+++ b/src/components/DeploymentsModal.tsx
@@ -1,3 +1,4 @@
+import './DeploymentsModal.css';
 import { fetchDeployments, stopDeployment } from '../api';
 import { CustomExclamationTriangleIcon } from './Icons';
 import { useDeploymentStore, useSettingsStore } from '@kaoto/store';
@@ -15,6 +16,11 @@ import {
   Title,
   AlertVariant,
   Badge,
+  Toolbar,
+  ToolbarContent,
+  ToolbarItem,
+  InputGroup,
+  SearchInput,
 } from '@patternfly/react-core';
 import { CubesIcon, HelpIcon } from '@patternfly/react-icons';
 import {
@@ -29,7 +35,7 @@ import {
   ActionsColumn,
 } from '@patternfly/react-table';
 import { useAlert } from '@rhoas/app-services-ui-shared';
-import { useEffect, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 
 export interface IDeploymentsModal {
   handleCloseModal: () => void;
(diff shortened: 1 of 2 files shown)


src/components/DeploymentsModal.tsx
```diff
@@ -185,54 +190,76 @@ export const DeploymentsModal = ({ handleCloseModal, isModalOpen }: IDeployments
...
-                    <Badge key={rowIndex} isRead>
-                      {dep.type}
-                    </Badge>
-                  </Td>
-                  <Td isActionCell>
-                    <ActionsColumn items={defaultActions(dep)} />
-                  </Td>
+          <>
+            <Toolbar>
+              <ToolbarContent className={'deoployments__toolbarContent'}>
```
kahboom, line 195: I think there's a typo in the className

kahboom: but I guess do we also need a classname here? I didn't see this added to the CSS

src/components/DeploymentsModal.tsx
```diff
@@ -130,6 +125,16 @@ export const DeploymentsModal = ({ handleCloseModal, isModalOpen }: IDeployments
     columnIndex,
   });
 
+  function search(items: any[]) {
```
kahboom, line 128: Probably we can assume that the search items will just be an array of strings, right? so maybe for the type let's put `items: string[]`, unless it gives you issues.

kahboom (changes requested):
@tplevko this looks great, thanks so much for your contributions. I added some comments, asking for very minor changes, but otherwise I'll be happy to merge.

src/components/DeploymentsModal.tsx
tplevko: this is an array of IDeployment's, so to make the function more specific changed to this type

src/components/DeploymentsModal.tsx
tplevko: sure, sounds good - removed. Thanks! 
Also removed the className={'table-wrap'} from the TableComposable, as I missed that one too.

kahboom (approved):
thank you @tplevko ! 🎉

PR #1200, opened 2023-02-01
fix(CodeEditor): Code sync when different DSL is set

### LINKED ISSUE #950: Code Editor: Update DSL in Settings automatically if doesn't match with DSL in code
## Background
We need to handle when the source code DSL doesn't match with the DSL in the Settings. For example, if the code has a DSL of Kamelet but in the Settings it's 'Camel Routes', it would return a blank page. This was partially fixed through the backend, but the frontend needs to handle this better.

Discussing with the team, we considered showing a warning to the user or automatically updating the Settings DSL, and decided on automatically updating the DSL in Settings to reflect the DSL from the code.

## Task
When the source code DSL doesn't match the DSL in Settings, automatically update the DSL in Settings, similarly to how we do for the Name property.

Relates to #787

### LINKED ISSUE #1121: When uploading a file, if the kind is not matching, the content is wiped out
### Describe the Bug

Kind is not adapted to uploaded files (also for edited files) and the content is wiped out without warning

_No response_

- Open Kaoto UI
- In the three vertical dots -> Settings, check the "Type" (by default "KameletBinding") then Cancel
- Click on the button to show the source
- Upload a file which is of kind integration, for instance:
```
apiVersion: camel.apache.org/v1
kind: Integration
metadata:
  name: integration
spec:
  flows:
  - from:
      uri: activemq:null:queue
      steps:
      - aggregate: {}
      - to:
          uri: log:null
```
- The source code is showing this code but the canvas stays empty
- when closing and reopening source, the content is wiped out and replaced by an empty KameletBinding

### Screenshots or Videos

wipedOutContent

### Platform

- OS: Fedora
- Browser: Chrome
- Version: main

mmelko:
- add dsl field into IIntegration interface to reflect backend model
- add aditional check for DSL in useEffect method of SourceCodeEditor
- fix #1121
- fix #950

### DIFF
--- a/src/components/SettingsModal.tsx
+++ b/src/components/SettingsModal.tsx
@@ -36,7 +36,7 @@ export const SettingsModal = ({ handleCloseModal, isModalOpen }: ISettingsModal)
   const [availableDSLs, setAvailableDSLs] = useState<string[]>([]);
   const { settings, setSettings } = useSettingsStore((state) => state);
   const [localSettings, setLocalSettings] = useState<ISettings>(settings);
-  const { integrationJson } = useIntegrationJsonStore((state) => state);
+  const { integrationJson, updateIntegration } = useIntegrationJsonStore((state) => state);
   const { setSourceCode } = useIntegrationSourceStore();
   const previousIntegrationJson = usePrevious(integrationJson);
   const previousName = usePrevious(localSettings.name);
@@ -99,6 +99,8 @@ export const SettingsModal = ({ handleCloseModal, isModalOpen }: ISettingsModal)
       capabilities.dsls.forEach((dsl) => {
         if (dsl.name === value) {
           setLocalSettings({ ...localSettings, dsl: dsl });
+          const tmpIntegration = { ...integrationJson, dsl: dsl.name };
+          updateIntegration(tmpIntegration);
           return;
         }
       });
--- a/src/components/SourceCodeEditor.tsx
+++ b/src/components/SourceCodeEditor.tsx
@@ -42,7 +42,6 @@ const SourceCodeEditor = (props: ISourceCodeEditor) => {
             const tmpSettings = { ...settings, dsl: dsl };
             setSettings(tmpSettings);
             fetchTheSourceCode(tmpInt, tmpSettings);
-            return;
           }
         });
       });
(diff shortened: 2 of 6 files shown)


src/components/SettingsModal.tsx
```diff
@@ -99,6 +99,9 @@ export const SettingsModal = ({ handleCloseModal, isModalOpen }: ISettingsModal)
       capabilities.dsls.forEach((dsl) => {
         if (dsl.name === value) {
           setLocalSettings({ ...localSettings, dsl: dsl });
+          const tmpIntegration = integrationJson;
+          tmpIntegration.dsl = dsl.name;
```
kahboom, line 103: We have to start trying to use the spread operator to prevent mutation, I know I do it some other places, but we're gonna have to try. 😄  Instead of this try:
```
const tmpIntegration = {...integrationJson, dsl: dsl.name}
```

src/components/SourceCodeEditor.tsx
```diff
@@ -32,17 +32,37 @@ const SourceCodeEditor = (props: ISourceCodeEditor) => {
...
     if (previousJson === integrationJson) return;
-    let tmpInt = integrationJson;
-    tmpInt.metadata = { ...integrationJson.metadata, ...settings };
-    fetchIntegrationSourceCode(tmpInt, settings.namespace).then((newSrc) => {
+
+    if (tmpInt.dsl != null && tmpInt.dsl !== settings.dsl.name) {
+      fetchCapabilities().then((capabilities: ICapabilities) => {
+        capabilities.dsls.forEach((dsl) => {
+          if (dsl.name === tmpInt.dsl) {
+            tmpSettings.dsl = dsl;
```
kahboom, line 43: Same here as above

src/components/SourceCodeEditor.tsx
```diff
@@ -32,17 +32,37 @@ const SourceCodeEditor = (props: ISourceCodeEditor) => {
...
+        });
+      });
+    } else {
+      fetchTheSourceCode(integrationJson, settings);
+    }
+  }, [integrationJson]);
+
+  const fetchTheSourceCode = (integration: IIntegration, settings: ISettings) => {
+    integration.metadata = { ...integrationJson.metadata, ...settings };
+    integration.dsl = settings.dsl.name;
```
kahboom, line 57: And here.. instead of `integration` below, do a new object like
```
{ ...integration, metadata: {...integrationJson.metadata, ...settings, dsl: settings.dsl.name} }
```
Of course test it first though because I'm just writing this on GitHub 😅

kahboom (changes requested):
Thanks @mmelko , I left some comments. Also there is a code smell reported that will need to be looked into. Tested locally and it looks fine otherwise.

kahboom (approved):
thanks @mmelko !

Issue #1154, opened 2023-01-24
Prefill default values in kamelet forms


tplevko:
### Please describe the feature that you want to propose

In Kamelet specifications there are also "default" values set for certain properties - most common db server ports. This prefilled values could provide some guidance for the users.

![Screenshot from 2023-01-24 10-29-59](https://user-images.githubusercontent.com/4180208/214256801-8208de8a-a297-44d5-9eec-9f9088a366bc.png)

Delawen:
Wouldn't it be better to have the source code and the forms with less information? Asking just because as a developer to me it makes sense to have a less verbose experience. But that may be me.

kahboom:
@Delawen we can always have a "clear" button too, to delete default values, because otherwise what would be the point of default values if we don't provide them?

Another benefit might be that we can use it for validation, as discussed in this PR: https://github.com/KaotoIO/kaoto-ui/pull/1216. So, for example, when one of the fields is required, right now we can only show an asterisk* to show it's required, but there's really no validation since we don't have a save button, however, if we fill the default values then we can highlight it in red if the user deletes those values, to show that they are required.

Tomas also mentioned the deploy button could also show the validation, if possible. We can discuss this with UX.

apupier:
With Kaoto backend 0.6.2, the path parameters are filled with null instead of the default values.

Delawen:
`, when one of the fields is required, right now we can only show an asterisk* to show it's required, but there's really no validation since we don't have a save button,`

Why don't we have a save button? On step extensions we have a save button. Why must be the default tab config different?

`because otherwise what would be the point of default values if we don't provide them?`

The default value is the value the integration framework will use if it comes empty from the flow definition.

kahboom:
>With Kaoto backend 0.6.2, the path parameters are filled with null instead of the default values.

@apupier - Do you know if an issue already exists in the backend for this?

>Why don't we have a save button?

@Delawen - Because when we first designed Kaoto we didn't have any kind of save function, everything was meant to be in sync automatically when the user types (same for the code editor). Now that the code editor has a sync button maybe it makes sense to change this.

>On step extensions we have a save button. Why must be the default tab config different? That must be confusing to the user, both default config and step extensions should play with the same rules.

Yes that's true, but I don't know why step extensions don't have a save button if the default config didn't, as that was created first.

>The default value is the value the integration framework will use if it comes empty from the flow definition.

We will need them for this enhancement and also for #1307.

[...]

PR #1216, opened 2023-02-06
fix(1153) Indicate which fields are required

### LINKED ISSUE #1153: Indicate which kamelet fields are required

The UI currently doesn't indicate for the user, which of the input fields in the specified kamelets are required:

Screenshot from 2023-01-24 09-13-42

This list however is present in the kamelet specification, e.g. for the above mentioned aws-redshift-source it's folowing list:

Screenshot from 2023-01-24 09-17-10

tplevko:
fixes https://github.com/KaotoIO/kaoto-ui/issues/1153

The required fields are highlighted with (*) 

Screenshot from 2023-02-06 11-44-03

### DIFF
--- a/src/components/VisualizationStepViews.tsx
+++ b/src/components/VisualizationStepViews.tsx
@@ -235,7 +235,7 @@ const VisualizationStepViews = ({
                 <Grid hasGutter style={{ maxHeight: 'calc(100vh - 300px)', overflow: 'auto' }}>
                   {step.parameters && (
                     <JsonSchemaConfigurator
-                      schema={{ type: 'object', properties: stepPropertySchema }}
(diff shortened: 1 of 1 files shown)


kahboom (approved):
thank you @tplevko !

I wish we could use the required field for validation, but without a save button I'm not sure it makes sense, especially since we are not currently filling in default values. But maybe with #1154 then we can use that to give users a hint? I guess it's something we can keep in mind for the future.

tplevko:
yes, the validation will be I think next step. I know the pushing of the deploy button could be a place where we could validate the filled forms, but with current flow I'm not sure how this would work.
This is probably an area for the UX team :)

Issue #1183, opened 2023-01-27
Refactor: Clearer separation of concerns


kahboom:
### Please describe the task that needs to be done

We recently took some steps to extract functions into a service where they could more easily be tested, however, there is still a lot of room for improvement. The components, in particular, should be refactored to be part of the presentation layer without any in-app state (where possible), and we can move the business logic into their own components. This will also make it much easier to bring back Storybook (#1100).

Similarly, the `visualizationService` contains a lot of business logic, where many of the functions are specific to the business logic, not just to the visualization. For example, `findStepIdxWithUUID` has nothing to do with the visualization, but it still part of the service. We should break really long file into two different ones with relevant concerns.

igarashitm:
I'm planning to introduce `stepService.tsx` and move internal step representation (IStepProps) handling from `visualizationService.tsx` to `stepService.tsx`. Naming aside :grin: Letting `visualizationService.tsx` focus on visualized steps (nodes/edges).
We could even wrap those service functions into a class - so those in `visualizationService.tsx` will be in `VisualizationService` class, and `StepService` class as well, so that the functions for same concern are bundled with a name. WDYT?

kahboom:
I think that sounds really good! The only part I'm not sure about is introducing classes into the code base, but it's because even though I think and organize the code in models I try to keep it very functional programming. To bundle together you can group them into an object too, like:

```js
const ExampleService = {
  exampleMethod: function(value) {
    // something here
  },

  anotherMethod: function(value) {
    // ...
  }
};

export default ExampleService;
```

And use it like:
```js
import ExampleService from "@kaoto/services";
ExampleService.exampleMethod('hello');
```

The only reason I had them as individual functions not wrapped is because I personally prefer a shorter way to access the method like we do now:
```js
import {
  exampleMethod
} from '@kaoto/services';
exampleMethod('hello');
```

But of course the problem is that then the imports can get very large, and you don't have them grouped or know which service it comes from. So, in that sense I think your approach makes more sense, because you are importing a service from `@kaoto/services` and it would be nice to see which one it is.

kahboom:
@igarashitm - Should this be closed now from https://github.com/KaotoIO/kaoto-ui/pull/1228?

igarashitm:
I'd like to keep it open until I finish the rest of the components, WDYT?

kahboom:
Ah yes, sorry, I got confused!

lhein:
lets move it to sprint 16 then or even later

PR #1228, opened 2023-02-09
fix(refactor): Clearer separation of concerns for visualization (#1183)

igarashitm:
* From now on the componets are expected to only initialize the stores and forward them directly to the services. In This way the store details, which often touches the business logic, are isolated from components and encapsulated into services
  * In this changeset, only Visualization.tsx and VisualizationStep.tsx are addressed. More will be expected later
* Turns services into classes and let them hold references to the stores, mainly for reducing the arguments on business logic methods invoked from components, therefore it helps more to decouple business logic from components. Also the related business logic routines are bundled together into a same namespace of class, which makes it easier to see where the method is declared and for what purpose.
* Splitted StepsService from VisualizationService, trying to isolate the logical model handling (e.g. IStepProps) from pure visualization handling (React Flow nodes and edges) as much as possible

### DIFF
--- a/src/components/PlusButtonEdge.tsx
+++ b/src/components/PlusButtonEdge.tsx
@@ -1,7 +1,7 @@
 import './PlusButtonEdge.css';
 import { MiniCatalog } from '@kaoto/components';
-import { findStepIdxWithUUID, insertableStepTypes, insertStep } from '@kaoto/services';
-import { useIntegrationJsonStore, useNestedStepsStore } from '@kaoto/store';
+import { StepsService, ValidationService } from '@kaoto/services';
+import {useIntegrationJsonStore, useNestedStepsStore, useVisualizationStore} from '@kaoto/store';
 import { IStepProps, IVizStepNode } from '@kaoto/types';
 import { Popover } from '@patternfly/react-core';
 import { PlusIcon } from '@patternfly/react-icons';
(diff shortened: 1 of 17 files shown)


kahboom:
@igarashitm there are a few code smells from Sonarcloud you might want to check out.

I was also going to suggest that we have consistent naming for the branch steps. For example, in one of my upcoming PRs I was going to rename `originStep` to `rootStep` or `rootNode`, since I think it's more understandable. But I guess it would make more sense to do from this PR. We could then document this in the code base through a README or similar (I definitely need to add all docs to the repo 😅).

Third thing is that I haven't had a chance to look at the alternatives to classes like with the context/provider pattern (sorry about that). Maybe I will get a chance to do that today. But if everything else works fine then I wonder if it makes sense to merge first and then do those changes after? My main concern is that the work I have to do today (adding branching visually) will have an insane amount of conflicts with your code base, but we can't afford regressions either, so it would require lots of hands on testing. 🤔

cc @lordrip @mmelko

igarashitm:
Filed #1233 to address the overhead concern on instantiating services. I'll look into the code smells from now :+1:

igarashitm:
Alright, addressed "code smells", but other than "@todo"s.

igarashitm:
@kahboom rebased on top of #1235 :+1:

src/components/PlusButtonEdge.tsx
```diff
@@ -1,7 +1,7 @@
 import './PlusButtonEdge.css';
 import { MiniCatalog } from '@kaoto/components';
-import { findStepIdxWithUUID, insertableStepTypes, insertStep } from '@kaoto/services';
-import { useIntegrationJsonStore, useNestedStepsStore } from '@kaoto/store';
+import { StepsService, ValidationService } from '@kaoto/services';
+import {useIntegrationJsonStore, useNestedStepsStore, useVisualizationStore} from '@kaoto/store';
```
kahboom, line 4: Make sure you have Prettier formatting set up in your IDE (see below for example on WebStorm).. though I might just add this to part of the CI workflow, or to automatically format the code on push.

Screen Shot 2023-02-10 at 3 14 04 pm

kahboom (approved):

kahboom:
@igarashitm - I forgot to ask you to put the service methods in ABC order for easier finding 😂  yes I'm crazy.. maybe we can do it in a follow-up PR next week.. it just makes it easy to look for stuff especially if you don't know the name of it.

igarashitm:
oh OK, I'll sort them in the next batch 👍

PR #1260, opened 2023-02-14
chore(service): Sort service methods in alphabetical order

igarashitm:


igarashitm:
@kahboom do we want to sort all methods (static/non-static) in alphabetical order like this?

kahboom (approved):
@igarashitm yes exactly! I know it may seem silly but it makes me happy. :D

igarashitm:
Nah it's fair preference, and the originator has to be on first priority to set the preference :)

PR #1270, opened 2023-02-16
feat(viz): show prepend step button if previous step has branches

### LINKED ISSUE #1162: Branching: Add ability to insert step after a step that contains branches

At the moment, it's not possible to insert a step between a step that contains branches, and the next step.

In the example below, it would be the equivalent of inserting a step between the root `filter` step and `kamelet:sink` (but not _within_ the `filter` branch):

Screen Shot 2023-01-19 at 11 41 54 am copy 4

In essence, this would be the same as prepending a step (similarly to how we are intending to do for branch steps, #1161 ), for the `kamelet:sink` step, and so should only be necessary for this kind of step (the step _after_ a step that contains branches), otherwise you would normally use `insertStep`.

Relates to #140 and should consider what happens visually when groups are collapsed (#772).

### Tasks
- [x] Make prepend step show up for steps whose previous step contains branches
- [x] Check for steps within branches with the same conditions
- [x] Handle prepending for both scenarios

kahboom:
This PR adds a prepend button to steps where its previous step has branches. Resolves #1162.

Kapture 2023-02-16 at 16 19 06

### DIFF
--- a/src/components/VisualizationStep.tsx
+++ b/src/components/VisualizationStep.tsx
@@ -22,6 +22,7 @@ const VisualizationStep = ({ data }: NodeProps<IVizStepNodeData>) => {
   const nestedStepsStore = useNestedStepsStore();
   const visualizationStore = useVisualizationStore();
   const integrationJsonStore = useIntegrationJsonStore();
+  const visualizationService = new VisualizationService(integrationJsonStore, visualizationStore);
   const stepsService = new StepsService(integrationJsonStore, nestedStepsStore, visualizationStore);
 
   const { addAlert } = useAlert() || {};
@@ -96,7 +97,7 @@ const VisualizationStep = ({ data }: NodeProps<IVizStepNodeData>) => {
           data-testid={`viz-step-${data.step.name}`}
         >
           {/* PREPEND STEP BUTTON */}
-          {VisualizationService.showPrependStepButton(data, endStep) && (
+          {visualizationService.showPrependStepButton(data) && (
             <Popover
               appendTo={() => document.body}
               aria-label="Search for a step"
--- a/src/services/stepsService.ts
+++ b/src/services/stepsService.ts
@@ -336,7 +336,8 @@ export class StepsService {
         const newRootStep = setDeepValue(stepCopy, pathToBranch, newBranch);
         this.integrationJsonStore.replaceStep(newRootStep, rootStepIdx);
       } else {
-        this.integrationJsonStore.appendStep(step);
+        const currentStepIdx = this.findStepIdxWithUUID(currentStep.UUID);
+        this.integrationJsonStore.prependStep(currentStepIdx, step);
       }
     });
   }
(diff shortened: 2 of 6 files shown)


src/components/VisualizationStep.tsx
```diff
@@ -96,7 +96,7 @@ const VisualizationStep = ({ data }: NodeProps<IVizStepNodeData>) => {
           data-testid={`viz-step-${data.step.name}`}
         >
           {/* PREPEND STEP BUTTON */}
-          {VisualizationService.showPrependStepButton(data, endStep) && (
+          {VisualizationService.showPrependStepButton(data, endStep, visualizationStore.nodes) && (
```
igarashitm, line 99: This could be changed to an instance method `visualizationService.showPrependStepButton(data)` then the method can check if it's endStep, and retrieve the `visualizationStore.nodes` from its reference, WDYT?

kahboom: Thanks @igarashitm , some questions:
1. Do you mean to instead check the `node.data.step.type` directly from `showPrependStepButton()`, basically what the `StepsService.isEndStep()` method does, but manually? 
2. If I remove the `visualizationStore.nodes` parameter how would I access the `visualizationStore`? It always tells me it doesn't exist on `VisualizationService`.

igarashitm (dismissed):
Only one minor comment, otherwise LGTM :+1:

src/components/VisualizationStep.tsx
igarashitm: 1. I think it's even better if `showPrependStepButton()` could just rely on `StepsService.isEndStep()`
2. ah so this component doesn't have the VisualizationService instance, I would do `const visualizationService = new VisualizationService(...)` just like this - https://github.com/KaotoIO/kaoto-ui/pull/1270/files#diff-714a52075514a965fd07e5a9d53ace053b12ff84b11f2c91f9041aa1124c3e3cL25

igarashitm: Acutually like this - https://github.com/KaotoIO/kaoto-ui/blob/main/src/components/Visualization.tsx#L42

igarashitm: and don't forget to remove `static` from `showPrependStepButton()`

kahboom: ahh how silly, I didn't see we had no instantiated it already from the component. Thank you, these are good ideas, I'll make the changes now.

kahboom: oops only one question--how can I instantiate the service from the unit tests? then we would have to mock the store as well, right?

kahboom: and then it complains because obviously I'm not using the `visualizationService` value, though maybe I don't need to assign it to anything and just do `new visualizationService()`, and unit tests cannot access if it's not `static`.

igarashitm: This is for `StepsService` though, I could instantiate with the real store (but empty) - https://github.com/KaotoIO/kaoto-ui/blob/main/src/services/stepsService.test.ts#L9-L13

igarashitm (approved):

PR #1276, opened 2023-02-17
Place the tooltips in code editor to top

tplevko:
The current - right placement of the tooltip was causing sometimes issues. Do you think that placing the tooltips to the top makes sense?

Recordings with right placement:
Peek 2023-02-17 13-36

top placement:
Peek 2023-02-17 13-38


kahboom (approved):
much better, thanks @tplevko , I'm not sure why we had it in that position before.

PR #1305, opened 2023-02-22
fix: Use "action" instead of "middle" for consistency

### LINKED ISSUE #1262: UX: Make messaging of steps consistent in notifications

The toast notification when dropping a `start` step onto an empty branch refers to a `middle` or `end` step, but the filter buttons in the catalog are `actions` and `end`. We should be using consistent messaging to not confuse users, so the notification should be using the same terminology as the catalog.

Validation messaging is likely located in `ValidationService`.

Screen Shot 2023-02-15 at 1 54 26 pm

Delawen:
Fixes #1262 

image


kahboom (approved):
thank you @Delawen !! <3

PR #1309, opened 2023-02-22
feat(viz): highlight connected branch nodes

### LINKED ISSUE #1133: Visualization: Show when steps are part of a branch

kahboom:
This PR adds the enhancement to show which steps are part of a branch. All nodes will now have a grey border that will turn blue on hover. For nodes/steps that contain branching, it will highlight the step and its child steps. Resolves https://github.com/KaotoIO/kaoto-ui/issues/1133.

## Screenshots

Kapture 2023-02-22 at 17 19 37

### DIFF
--- a/src/components/Visualization.css
+++ b/src/components/Visualization.css
@@ -20,7 +20,7 @@
 .stepNode {
     align-items: center;
     background: var(--pf-global--BackgroundColor--100);
-    border: 2px solid rgb(0, 136, 206);
+    border: 2px solid var(--pf-global--BorderColor--100);
     border-radius: 50%;
     display: block;
     font-family: Ubuntu, serif;
@@ -30,6 +30,10 @@
     padding: 10px;
     text-align: center;
     width: 80px;
+}
+
+.stepNode__Active {
+    border: 2px solid rgb(0, 136, 206);
 }
 
 .stepNode__Add {
--- a/src/components/VisualizationStep.tsx
+++ b/src/components/VisualizationStep.tsx
@@ -26,6 +26,7 @@ const VisualizationStep = ({ data }: NodeProps<IVizStepNodeData>) => {
   const stepsService = new StepsService(integrationJsonStore, nestedStepsStore, visualizationStore);
   const showBranchesTab = VisualizationService.showBranchesTab(data);
   const showStepsTab = VisualizationService.showStepsTab(data);
+  const supportsBranching = StepsService.supportsBranching(data.step);
 
   const { addAlert } = useAlert() || {};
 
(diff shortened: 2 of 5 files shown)


Delawen:
[Screencast from 2023-02-23 10-01-32.webm](https://user-images.githubusercontent.com/726590/220862617-6b89d88b-b5e4-4675-b05d-6cb001d3f5ba.webm)

Nested branches don't highlight their inner branches. Is that expected?

kahboom:
@Delawen unfortunately yes, there is no distinction for the nested branches (yet), I discussed this with UX and this is essentially a first pass to improve the existing UX for branching.

Delawen (approved):

PR #1315, opened 2023-02-23
feat(branching): add support for min/max branches

### LINKED ISSUE #1143: Add support for minBranches/maxBranches

Some EIP steps require a minimum number of branches, so, once we support adding branches visually, we should verify that the minimum and maximum number of branches are correct.

## Requirements
1. If the user does not have enough branches, we should create the branches for the user automatically either via the backend or frontend. Alternatively, we can simply flag it to the user either in the visualization or in a step extension (preferably in the visualization).
2. If the user has the maximum number of branches, the button to add more branches should be hidden or disabled.

## Tasks

- [x] Hide `delete` button for edge for x number of edges if the step has a minimum number of branches
- [x] Hide branches tab if the step has the same amount as max branch
- [ ] Hide plus button if the step doesn't support branching and has already reached the maximum number of branches? Would that confuse users? Or maybe I can just disable it..
- [x] Custom tooltip message for mini catalog branches tab
- [ ] Auto-create minimum number of required branches?

requires #786

kahboom:
This PR adds support for handling minimum and maximum branches.
Resolves #1143 

## Changes
- Add methods for checking if branches have reached the minimum threshold and maximum number of branches
- Hide button to delete branch if deleting will drop number of branches below the minimum
- Disable Branches tab in mini catalog if max number of branches have been reached
- Add custom validation tooltip message for when steps have reached their maximum number of branches, to explain why the branches tab is disabled
- Rename `originStep` and `parentStep` to `rootStep` for improved clarity
- Update services `README`
- Add a sneaky `codecov.yml` file

## Screenshots

**NOTE:** In the following examples, `filter` has a minimum number of branches of 1 and a maximum number of branches of 1.

Minimum threshold has not been met (`minBranches` + 1), so delete branch button is hidden, but append step shows because you can still add a step after the branch:

Show append step button

Once the maximum number of branches is reached, disable the Branches tab:
Disable Branches tab

On reaching minimum number of branches, show delete branch button again (this is just for an example, `filter` max number of branches is usually 1):

Show delete branch button

Once you have a next step, you can no longer add a step, only branches, but in this case the maximum number of branches has also been reached, so you can't add either. In this case we hide the append step button:

Hide append step button

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,8 @@
+coverage:
+  status:
+    project:
+      default:
+        # basic
+        target: auto
+        threshold: 0.5%
+        base: auto
--- a/src/components/VisualizationStep.tsx
+++ b/src/components/VisualizationStep.tsx
@@ -207,12 +207,16 @@ const VisualizationStep = ({ data }: NodeProps<IVizStepNodeData>) => {
           {VisualizationService.showAppendStepButton(data, endStep) ? (
             <Popover
               appendTo={() => document.body}
-              aria-label="Search for a step"
+              aria-label="Add a step or branch"
               bodyContent={
                 <MiniCatalog
                   children={<BranchBuilder handleAddBranch={handleAddBranch} />}
                   disableBranchesTab={!showBranchesTab}
-                  disableBranchesTabMsg={"This step doesn't support branching."}
+                  disableBranchesTabMsg={ValidationService.getBranchTabTooltipMsg(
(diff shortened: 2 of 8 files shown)


lordrip (approved):
Tested locally, LGTM.

We would need to check with @Delawen if there's a possibility to know the branch identifiers alongside the newly added branches, otherwise, we would need to add to the Circuit Breaker extension (among others perhaps) a way to rename the branches.

Issue #1233, opened 2023-02-10
Refactor: Investigate any better way for managing services lifecycle


igarashitm:
### Please describe the task that needs to be done

As a part of #1183 work, #1228 is turning services into class, and components are going to instantiate them like this
https://github.com/KaotoIO/kaoto-ui/pull/1228/commits/b3946131e6d8861c46d8724e50961272ed54d356#diff-b05d69f0aef9ae8d739a7510b44be6747508e1709033a2e44f609523c404f5fcR46

One concern here is the instantiation happens on every rerendering, on every component. It might cause performance degradation. Investigate if there's any better way. Possible option is context/provider pattern.

igarashitm:
cf. https://github.com/KaotoIO/kaoto-ui/pull/1338/files#diff-20fc6180949d2fcf17d991cc62c6be55e572335ea16c00c805dcf3c7a8a6089dR41

PR #1338, opened 2023-02-28
fix(vscode): Adding a step in Kaoto VSCode takes several seconds

### LINKED ISSUE #132: Adding a step can take more than 12 seconds
seems to be since backend 0.6.2 (to be confirmed)

image

image

lordrip:
### Description
Currently, adding a new step from the `VSCode` extension takes more than 12 seconds.

This is because, in `VSCode`, the `MiniCatalog` component gets rerendered many times blocking the UI thread.

### Examples
* KaotoUI - Firefox Developer edition
Screenshot from 2023-02-28 15-02-26

* KaotoUI - VSCode
Screenshot from 2023-02-28 15-04-50

Notice how in VSCode there are many more render passes hence blocking the UI thread.

Part of the solution is to avoid recreating functions inside of components that later on are passed to child components. A more robust solution would include removing many other points that trigger extra renders.

### Changes
* Move the `onNodeClick` function inside of a `useCallback()` hook
* Given that stepsService it's a dependency of `onNodeClick`, it needed to be moved inside a `useMemo()` hook.
* Remove `setReactFlowInstance` since seems not used
* Remove `reactFlowWrapper` since is not linked with anything
* Move inline styles to a dedicated CSS class

Fixes https://github.com/KaotoIO/vscode-kaoto/issues/132

### DIFF
--- a/src/components/Visualization.tsx
+++ b/src/components/Visualization.tsx
@@ -11,7 +11,7 @@ import {
 import { StepsService, VisualizationService } from '@kaoto/services';
 import { useIntegrationJsonStore, useNestedStepsStore, useVisualizationStore } from '@kaoto/store';
 import { IStepProps, IVizStepNode } from '@kaoto/types';
-import { useEffect, useMemo, useRef, useState } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import ReactFlow, { Background, Viewport } from 'reactflow';
 
 const Visualization = () => {
@@ -25,8 +25,6 @@ const Visualization = () => {
     zoom: 1.2,
   };
   const [isPanelExpanded, setIsPanelExpanded] = useState(false);
-  const [, setReactFlowInstance] = useState(null);
-  const reactFlowWrapper = useRef(null);
   const [selectedStep, setSelectedStep] = useState<IStepProps>({
     maxBranches: 0,
     minBranches: 0,
@@ -40,7 +38,10 @@ const Visualization = () => {
   const previousLayout = useRef(visualizationStore.layout);
   const previousIntegrationJson = useRef(integrationJsonStore.integrationJson);
   const visualizationService = new VisualizationService(integrationJsonStore, visualizationStore);
-  const stepsService = new StepsService(integrationJsonStore, nestedStepsStore, visualizationStore);
+  const stepsService = useMemo(
+    () => new StepsService(integrationJsonStore, nestedStepsStore, visualizationStore),
+    [integrationJsonStore, nestedStepsStore, visualizationStore],
+  );
 
   // initial loading of visualization steps
   useEffect(() => {
@@ -109,7 +110,7 @@ const Visualization = () => {
    * @param e
    * @param node
    */
-  const onNodeClick = (e: any, node: IVizStepNode) => {
+  const onNodeClick = useCallback((e: any, node: IVizStepNode) => {
     // here we check if it's a node or edge
     // workaround for https://github.com/wbkd/react-flow/issues/2202
     if (!e.target.classList.contains('stepNode__clickable')) return;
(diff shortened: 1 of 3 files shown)


src/components/Visualization.tsx
```diff
@@ -25,8 +25,6 @@ const Visualization = () => {
     zoom: 1.2,
   };
   const [isPanelExpanded, setIsPanelExpanded] = useState(false);
-  const [, setReactFlowInstance] = useState(null);
```
lordrip, line 28: @kahboom is this needed? I think that it forces a render but I don't know if there another use case for it :thinking:

src/components/Visualization.tsx
```diff
@@ -162,19 +159,14 @@ const Visualization = () => {
...
+        position="right"
+        id="right-resize-panel"
+        defaultSize="500px"
+        minSize="150px"
       >
         <div
           className="reactflow-wrapper"
-          data-testid={'react-flow-wrapper'}
-          ref={reactFlowWrapper}
-          style={{
```
lordrip, line 174: Moved to the corresponding `Visualization.css` file

igarashitm (dismissed):

src/components/Visualization.tsx
```diff
@@ -40,7 +38,10 @@ const Visualization = () => {
   const previousLayout = useRef(visualizationStore.layout);
   const previousIntegrationJson = useRef(integrationJsonStore.integrationJson);
   const visualizationService = new VisualizationService(integrationJsonStore, visualizationStore);
-  const stepsService = new StepsService(integrationJsonStore, nestedStepsStore, visualizationStore);
+  const stepsService = useMemo(
```
igarashitm, line 41: :eyes: just curious, should `useMemo()` be the default way of instantiating (and caching) service instances? a "memo" for #1233

apupier (commented):
Works great in VS Code. Not detected issues. I really do not know the potential of the code changes so I only provide a comment not approve.

src/components/Visualization.tsx
lordrip: That's a very interesting question :smiley: 

From my point of view, I don't think that this is the definite way to go but a transition point.

At first, the `onNodeClick` function was being recreated on every render pass, together with the `stepService`, so as I wrapped `onNodeClick` function into a `useCallback()` hook to avoid recreating it, `stepService` now it's a required dependency of it, so I wrapped the service with `useMemo()` to recreate it only if something change.

I think that a more appropriate fix would be to have this service [as a context](https://beta.reactjs.org/reference/react/useContext), this way we wouldn't need to create the service in different places but just call the `useContext()` hook where needed.

igarashitm: aha! makes sense, thanks :+1:

src/components/Visualization.tsx
kahboom: confession time--I have no idea. I think that might have been from an older version of React Flow, but I can't be sure. 😅  We can remove it, just be sure to check rendering still happens as needed with the basic actions (for example replacing from big catalog, deleting steps, editing config, etc.)

kahboom (dismissed):
@lordrip It looks good to me. Did some basic testing locally. Will wait on you before merging, since I'm not sure if you still wanted to remove the `setReactFlowInstance` code, particularly if it is causing an extra render we don't need.

src/components/Visualization.tsx
lordrip: Mmm, I think that you're right, I checked [here](https://reactflow.dev/docs/guides/migrate-to-v10/#4-onload---oninit) and it seems that it's also deprecated.

lordrip:
> @lordrip It looks good to me. Did some basic testing locally. Will wait on you before merging, since I'm not sure if you still wanted to remove the `setReactFlowInstance` code, particularly if it is causing an extra render we don't need.

Yes, it seems that `onLoad` is deprecated and renamed to `onInit` instead, and seems that we're not using it either.

PR #1375, opened 2023-03-03
Add gzip config to nginx server

### LINKED ISSUE #1363: Enable gzip on nginx proxy
### Please describe the task that needs to be done

At the moment, gzip is not enabled in the [nginx config](https://github.com/KaotoIO/kaoto-ui/blob/main/nginx/nginx.conf), and this affects how much data is transferred over the wire, especially for the catalog.

The goal it's to enable `gzip` and `brotli` compression algorithms to reduce the amount of data being transferred.

lordrip:
### Description
At the moment, there is no gzip configuration enabled for the `nginx` server, meaning that all payloads are being sent to clients without compression.

This commit enabled gzip on the `nginx` side to reduce the amount of data transferred over the wire

### Changes
* Enable `gzip`
* Format the `nginx.conf` file using the VS Code [NGINX Configuration Language Support](https://marketplace.visualstudio.com/items?itemName=ahmadalli.vscode-nginx-conf)

| filename | size | gzip |
| --- | --- | --- |
| mini-catalog | 1.77 MB | 37.94 kB |
| catalog | 7.18 MB | 133.73 kB |

### Before
Screenshot from 2023-03-03 11-42-31

### After
Screenshot from 2023-03-03 11-47-48

fixes https://github.com/KaotoIO/kaoto-ui/issues/1363

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -3,57 +3,63 @@ worker_processes auto;
 error_log /var/log/nginx/error.log notice;
 pid /tmp/nginx.pid;
 
-events {
-  worker_connections 1024;
+events
+{
+	worker_connections 1024;
 }
 
-http {
-  proxy_temp_path /tmp/proxy_temp;
-  client_body_temp_path /tmp/client_temp;
-  fastcgi_temp_path /tmp/fastcgi_temp;
-  uwsgi_temp_path /tmp/uwsgi_temp;
-  scgi_temp_path /tmp/scgi_temp;
+http
+{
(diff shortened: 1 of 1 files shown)


mmelko (commented):
looks good but I don't think this is a correct formatting: 
please use https://nginxbeautifier.github.io/ or something similar

mmelko (changes requested):
see the comment above

kahboom (approved):

mmelko (approved):

Issue #1383, opened 2023-03-03
Add possibility to sync cleared code editor with canvas


tplevko:
### Please describe the feature that you want to propose

This is more a question regarding the handling of when user clears the code editor. There is currently no way for the users to sync a clear editor with the canvas, the "sync" button is disabled for clean canvas.
I know, there is in that case nothing to sync, as there is nothing in the editor - but on the other hand, in case users insert invalid integration definition into the editor and tries to sync, it's updated to a stub integration for them.
Shouldn't this be also an option in case the code editor is empty?

Current state recording for reference:

![Peek 2023-03-03 16-44](https://user-images.githubusercontent.com/4180208/222764346-c791983d-9e7a-4a83-bf3f-92dc8d57a12f.gif)

kahboom:
Thanks @tplevko . Just to be clear, are you saying that when you press the clear button that it should also update the canvas to remove the nodes? If so, I agree and I even suggest this is a bug.

Or are you saying that we should allow clearing the code editor without it returning the stub/empty integration code like it does now?

tplevko:
Hi @kahboom - I wanted to say the first thing you wrote. I think with the "hanging" integration in canvas, it could be confusing for the users, whether the "clean" action was completed successfully.
 
I don't have anything against the stub/empty integration code. I think this is actually useful for the users.

Delawen:
Related to #1234

Issue #1297, opened 2023-02-21
E2E tests for critical user journeys


kahboom:
### Please describe the task that needs to be done

We can use E2E tests to test individual components visually as well as to test user journeys (traditional E2E).

For this task, we don't need to test the presence or appearance of individual components as much as the user journey behaviors/actions, and that the user sees what is expected after those actions. If possible, it would be nice to test that the edges and nodes also don't disappear after certain actions. 😂  (example bug: https://github.com/KaotoIO/kaoto-ui/issues/1288)

![Nodes and Edges](https://user-images.githubusercontent.com/3844502/220414598-7a6210fc-a2de-4d51-857c-2c9dcd7306e1.png)

# Scenarios
User journeys with each visual component, some testing a "bad" behavior (invalid) to see if user sees proper validation.

\* = lower priority issue

## Catalog actions
- [ ] User drags and drops an invalid step onto a nested branch step (i.e. a `start` step when branches should start with `action` steps)
- [ ] User replaces a placeholder step via drag & drop from the big catalog
- [ ] User replaces a placeholder step using the mini catalog*

## Code editor actions
- [ ] User adds, edits, and removes a step from the YAML*
- [ ] User adds and removes a branch from the YAML
- [ ] User undoes a change they saved, syncs with canvas*
- [ ] User clears the state of the code editor, syncs with canvas*
- [ ] User uploads YAML file, syncs with canvas
- [ ] User downloads YAML file, syncs with canvas

## Step actions from the canvas
- [ ] User inserts a step between two steps (+ button in between two nodes)
- [ ] In an integration with at least two steps, user deletes the first step, showing a placeholder step in its place
- [ ] User appends a step (+ button to the right of the node)
- [ ] User prepends a step (+ button to the left of the node)
- [ ] Step Detail - User configures a normal step, which updates the YAML
- [ ] Step Extensions
  - [ ] User sees a step extension (e.g. `choice` step)
  - [ ] User configures a step from a step extension
  - [ ] User adds and removes a branch from a step using a step extension

## Branching actions from the canvas
- [ ] User adds a branch from the canvas
- [ ] User deletes a branch from the canvas
- [ ] User completes normal actions on steps in a branch (configure, delete, replace, add)
- [ ] User appends a step in a branch (+ button to the right of the node, usually for the last step in a branch)
- [ ] User prepends a step in a branch (+ button to the left of the node, usually for the first step in a branch)
[...]

PR #1384, opened 2023-03-03
Add new e2e Kaoto editor tests

tplevko:
Added new Kaoto editor tests - related to https://github.com/KaotoIO/kaoto-ui/issues/1297

Code editor actions

- [x] User adds, edits, and removes a step from the YAML*
- [x] User adds and removes a branch from the YAML
- [x] User undoes a change they saved, syncs with canvas*
- [ ] User clears the state of the code editor, syncs with canvas*
- [x] User uploads YAML file, syncs with canvas
- [ ] User downloads YAML file, syncs with canvas

The item **User clears the state of the code editor, syncs with canvas** is blocked ATM by https://github.com/KaotoIO/kaoto-ui/issues/1383
The second missing item @kahboom -  **User downloads YAML file, syncs with canvas** - I may have not understood this correctly, but the download doesn't affect the editor, so is this test necessary?

### DIFF
--- a/cypress/e2e/8-code-editor-actions/code_editor_actions.cy.js
+++ b/cypress/e2e/8-code-editor-actions/code_editor_actions.cy.js
@@ -1,0 +1,110 @@
+describe('editing properties', () => {
+  beforeEach(() => {
+    let url = Cypress.config().baseUrl;
+
+    cy.visit(url);
+
+    cy.get('[data-testid="toolbar-show-code-btn"]').click();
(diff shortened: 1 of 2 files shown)


kahboom:
>The second missing item @kahboom - User downloads YAML file, syncs with canvas - I may have not understood this correctly, but the download doesn't affect the editor, so is this test necessary?

@tplevko that can be ignored/removed, I think I have just included this for completeness and maybe manual testing 🙂

kahboom (approved):

PR #1393, opened 2023-03-06
fix(jest): Jest warning about duplicated step mock

lordrip:
By adding a step.ts mock file, jest creates a warning because now there are 2 mock files with the same name after the build.

### Before
Screenshot from 2023-03-06 10-29-18

### After
Screenshot from 2023-03-06 10-29-39

This commit adds the dist folder to the jest config file.


unsortedhashsets (approved):
LGTM,

Assets are duplicating during `yarn build:lib` ->  `cpr ./src/assets ./dist/lib/assets`
The `/dist/` part is not taken into account in the coverage test and therefore can be excluded, I think.

kahboom (approved):
thanks @lordrip !

PR #1397, opened 2023-03-06
fix(viz): Avoid rerendering VisualizationStepViews component

### LINKED ISSUE #1314: Step extension gets rerendered upon hovering on a step
### Describe the Bug

When having an extension view open, hovering over a step from the canvas forces a rerender of the extension, causing that user data that has not been saved to get lost.

1. Set the type to `Integration`
2. Add any step, for instance, `activemq`
3. Add a `choice` step and click on the step to show the extension
4. Hover over the step and watch how the extension gets reloaded

### Screenshots or Videos

[Screencast from 2023-02-23 15-54-43.webm](https://user-images.githubusercontent.com/16512618/220942762-6a05bf2e-e8bc-47c6-99e4-73e808f20c1e.webm)

### Platform

- OS: [Proudly Fedora :heart:]
- Browser: [Chrome, Firefox]
- Version: [0.6.2-dev]

lordrip:
### Context
At the moment, whenever the step's detail view is open, hovering over a step forces the `VisualizationStepViews` component to rerender creating a content flash.

### Changes
Built on top of the previous commit, this commit removes the
need of binding entire stores in the following components:

- Visualization
- VisualizationStepViews

and only subscribes to the individual properties that need to be watched.

### Example

[Screencast from 2023-03-06 13-22-37.webm](https://user-images.githubusercontent.com/16512618/223109302-4e2c5e00-b9af-4303-b74c-35d4cf999caa.webm)

Fixes #1314

### DIFF
--- a/src/services/stepsService.ts
+++ b/src/services/stepsService.ts
@@ -7,7 +7,7 @@ import {
   startDeployment,
   stopDeployment,
 } from '@kaoto/api';
-import { IIntegrationJsonStore, INestedStepStore, RFState } from '@kaoto/store';
+import { IIntegrationJsonStore, INestedStepStore, RFState, useIntegrationJsonStore, useNestedStepsStore, useVisualizationStore } from '@kaoto/store';
 import {
   IIntegration,
   IKaotoApi,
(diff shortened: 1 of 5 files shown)


src/services/stepsService.ts
```diff
@@ -46,19 +52,19 @@ export class StepsService {
       } else {
         newStep.branches = newStep.branches.concat(branch);
       }
-      this.integrationJsonStore.replaceBranchParentStep(newStep, currentStepNested.pathToStep);
+      useIntegrationJsonStore.getState().replaceBranchParentStep(newStep, currentStepNested.pathToStep);
```
igarashitm, line 55: This looks promising, as it still helps to avoid too many method arguments that could be retrieved from the stores - which also means the components have to dig into stores and then the business logic leaks into components.

igarashitm: Actually it's better in a way that components don't have to pass the stores at all :+1:

kahboom (changes requested):
thanks @lordrip this looks good, just be sure to take a look at the sonarcloud code smell (it's a tiny one). it's complaining about a duplicate import (kind of surprised eslint didn't pick this up, maybe we don't have a rule for it): https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=1397&id=KaotoIO_kaoto-ui&open=AYa24eOH2Aww0WPj04MN

mmelko:
This look good! Definitely step to the right direction. 👍

kahboom (approved):

PR #1399, opened 2023-03-06
fix: Labels of CBR to branches connection get lost

### LINKED ISSUE #1279: Labels of CBR to branches connection get lost
### Describe the Bug

When creating a CBR and adding 2 when and 1 otherwise branch the otherwise label is visible for the connection to the otherwise branch. When you then set the simple condition for the 2 other branches and start to populate the branches with steps the otherwise label will disappear and it is not clear what makes the conditions for the other branches causing it to show up on the connection or not. I fear it will only be displayed if the length of the condition fits the width of the connection line but it is not always displayed.

1. set camel route type
2. choose a source of your choice
3. add a choice next to it
4. add 2 Whens and 1 otherwise via step extension
5. fill the conditions of the 2 when nodes
6. start populating the whens and the otherwise with at least 2 steps and monitor what happens to the connection labels

### Screenshots or Videos

_No response_

### Platform

- Version: [e.g. 0.30.0]

igarashitm:
Fixes: #1279

Custom Edge needs to put the label manually - https://reactflow.dev/docs/examples/edges/custom-edge/

### DIFF
--- a/src/components/DeleteButtonEdge.tsx
+++ b/src/components/DeleteButtonEdge.tsx
@@ -5,13 +5,15 @@ import { useIntegrationJsonStore, useNestedStepsStore, useVisualizationStore } f
 import { IVizStepNode } from '@kaoto/types';
 import { Button, Popover, Tooltip } from '@patternfly/react-core';
 import { MinusIcon } from '@patternfly/react-icons';
-import { getBezierPath, Position, useReactFlow } from 'reactflow';
+import { ReactNode } from 'react';
+import { EdgeText, getBezierPath, Position, useReactFlow } from 'reactflow';
 
 const foreignObjectSize = 40;
 
 export interface IDeleteButtonEdge {
   data?: any;
   id: string;
+  label?: string | ReactNode;
   sourceX: number;
   sourceY: number;
   targetX: number;
@@ -25,6 +27,7 @@ export interface IDeleteButtonEdge {
 /* PLUS BUTTON TO INSERT STEP */
 const DeleteButtonEdge = ({
   id,
+  label,
   sourceX,
   sourceY,
   targetX,
@@ -112,6 +115,13 @@ const DeleteButtonEdge = ({
           </Popover>
         </span>
       </foreignObject>
+      {label && (
+        <EdgeText
+          x={edgeCenterX + foreignObjectSize / 2}
+          y={edgeCenterY + foreignObjectSize / 2}
+          label={label}
+        />
+      )}
     </>
   );
 };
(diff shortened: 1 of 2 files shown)


src/components/DeleteButtonEdge.tsx
```diff
@@ -112,6 +114,13 @@ const DeleteButtonEdge = ({
           </Popover>
         </span>
       </foreignObject>
+      {label && (
+        <EdgeText
+          x={edgeCenterX + foreignObjectSize / 2}
```
igarashitm, line 119: To avoid overlaying the button, moving the label away from foreign object. Is there a better way to layout it?

kahboom: This is not so different to how we used to space out the steps, so if there is a better way then I'm not aware. :D I do wonder how much we would be able to integrate ELK with the labels, right now it's 100% handled by React Flow. ELK has a concept of labels, but I'm not sure what those labels mean exactly and if it's the same thing: https://www.eclipse.org/elk/reference/options/org-eclipse-elk-labelManager.html

kahboom (approved):

src/components/DeleteButtonEdge.tsx
```diff
@@ -5,13 +5,15 @@ import { useIntegrationJsonStore, useNestedStepsStore, useVisualizationStore } f
...
-import { getBezierPath, Position, useReactFlow } from 'reactflow';
+import { ReactNode } from 'react';
+import { EdgeText, getBezierPath, Position, useReactFlow } from 'reactflow';
 
 const foreignObjectSize = 40;
 
 export interface IDeleteButtonEdge {
   data?: any;
   id: string;
+  label?: string | ReactNode;
```
lordrip, line 16: If I'm not mistaken, the `ReactNode` type already include `string` in its definition, so maybe is not needed to explicitly pass it here.

```typescript
    type ReactNode = ReactElement | string | number | ReactFragment | ReactPortal | boolean | null | undefined;
```

```suggestion
  label?: ReactNode;
```

src/components/PlusButtonEdge.tsx
```diff
@@ -5,13 +5,15 @@ import { useIntegrationJsonStore, useNestedStepsStore, useVisualizationStore } f
...
-import { getBezierPath, Position, useReactFlow } from 'reactflow';
+import { ReactNode } from 'react';
+import { EdgeText, getBezierPath, Position, useReactFlow } from 'reactflow';
 
 const foreignObjectSize = 40;
 
 export interface IPlusButtonEdge {
   data?: any;
   id: string;
+  label?: string | ReactNode;
```
lordrip, line 16: maybe the same here :)

```suggestion
  label?: ReactNode;
```

lordrip (approved):

src/components/DeleteButtonEdge.tsx
igarashitm: ugh I missed this one, and right, just `ReactNode` was enough - will follow up. I blame React Flow as this is just a stolen one from its source :grin:

Issue #1277, opened 2023-02-17
Branching: Delete branch button not showing for empty & single-step branches


kahboom:
### Describe the Bug

There is no way to delete branches that only contain a single step or that are empty with a placeholder. This is likely a regression.

It's unclear if it's simply a matter of the button being hidden (since there is not much space, see #1146), but it's likely that the delete button is simply not rendering for that type of edge. There is a function dedicated to building single-step edges, called `buildBranchSingleStepEdges`, so this is likely where the bug is.

### Steps to Reproduce the Bug or Issue

1. Open source code editor
2. Paste the following YAML into it
3. Press "Sync your code"
4. See empty branches or single-step branches (`filter` and `loop` for the YAML below)

```yaml
apiVersion: camel.apache.org/v1alpha1
kind: Kamelet
metadata:
  annotations:
    camel.apache.org/kamelet.icon: cherry
  labels:
    camel.apache.org/kamelet.type: action
  name: eip-action
spec:
  definition:
    title: EIP Kamelet
    description: Used to test all EIP we implement
    properties: {}
  dependencies:
  - camel:core
  - camel:kamelet
  template:
    from:
      uri: kamelet:source
      steps:
      - loop:
          constant: '3'
          copy: true
          steps: []
      - choice:
          when:
          - simple: '{{?foo}}'
            steps:
            - to:
                uri: digitalocean:null
                parameters:
                  resource: '{}'
            - set-header:
                name: bar
                simple: foo
          - simple: '{{?bar}}'
            steps:
            - delay: {}
            - marshal:
                json:
                  library: Gson
          otherwise:
            steps:
            - pipeline: {}
            - claim-check:
                operation: Push
                key: foo
                filter: header:(foo|bar)
      - filter:
          simple: '{{?foo}}'
          steps: []
      - to:
          uri: kamelet:sink
```

### Screenshots or Videos

![Screen Shot 2023-02-17 at 12 54 04 pm](https://user-images.githubusercontent.com/3844502/219658549-e1a6d4a3-3a6c-4018-9355-baa28cb709b2.png)

### Platform

- OS: macOS, Linux
- Browser: Chrome
- Version: `0.6.2-dev`

Likely everywhere..

mmelko:
@kahboom I can see a root cause here:
https://github.com/KaotoIO/kaoto-ui/blob/9077f41f053675efb05e853da91f22fab973633e/src/services/validationService.ts#L175
[...]

PR #1407, opened 2023-03-07
fix(viz): change outgoing edge for placeholders to default

### LINKED ISSUE #1378: Delete empty branch two buttons - second one not working
### Describe the Bug

There are two buttons on connections going into an empty branch step and going from it. The direction from the empty branch step to end step is not working.

1. Crate empty branch from a step
2. Try to delete it using the "delete" button on the connection going from the step towards the end step - see it's not working

### Screenshots or Videos

Peek 2023-03-03 14-24

### Platform

- OS: [Linux]
- Browser: [Firefox]
- Version: [latest]

kahboom:
Fixes #1378, where a delete button on the right-hand side isn't working for empty branches. This delete button should never have been there to begin with. The delete button not showing on the left-hand side seems to be related to a separate issue, https://github.com/KaotoIO/kaoto-ui/issues/1277, but that button works as expected.

## Changes
- Changed the outgoing edge of a placeholder within a branch from `delete` to `default`
- Fixed expected output of edges unit test

## Screenshots
Screen Shot 2023-03-07 at 3 06 04 pm

### DIFF
--- a/src/services/visualizationService.test.ts
+++ b/src/services/visualizationService.test.ts
@@ -94,39 +94,39 @@ describe('visualizationService', () => {
       expect(
         VisualizationService.buildBranchSingleStepEdges(node, rootNode, rootNodeNext, 'CUSTOM-NODE')
       ).toEqual([
-          {
-            arrowHeadType: 'arrowclosed',
-            id: 'e-undefined>undefined',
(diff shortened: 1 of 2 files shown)


mmelko (approved):
LGTM 👍

PR #1430, opened 2023-03-09
Refactoring 09-step_actions_from_canvas

unsortedhashsets:
Hi!
A minor refactoring for 09-step_actions_from_canvas.
I will expand it on other tests if you are ok with that,
@kahboom @tplevko is it Ok?


kahboom (approved):

kahboom:
@unsortedhashsets is this okay to review / ready to merge? I approved only before I saw you removed request for review

unsortedhashsets:
@kahboom yes it is ready, I just wanted some review from your side or Tomas.
I want to make the same refactor for other parts of E2E

kahboom (approved):
i like the change! i'm not too familiar if this is expected practice from Cypress but I think it makes a lot of sense to avoid repeating ourselves in the tests. 👍

Issue #1280, opened 2023-02-17
wrong tab preselection in right panel


lhein:
### Describe the Bug

I created a Kamelet and checked each of the steps configuration. When clicking on each step the Config tab in the right panel is automatically selected but for the choice EIP it is not which is weird.

### Steps to Reproduce the Bug or Issue

Create a Kamelet as seen in the video and then select one node after another and check what tab is preselected.

### Screenshots or Videos

![wrong_tab_preselectio](https://user-images.githubusercontent.com/435853/219695739-c13a0ec2-5c36-4bb9-81db-97c0a2eed5dc.gif)

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

Delawen:
This is probably because the default `Config` tab is replaced on that case.

PR #1441, opened 2023-03-10
fix(defaultConfigTab): add setIndex to useEffect

### LINKED ISSUE #1433: Always select a tab by default in config panel

When switching between Step having a Step Extension and some that are relying on the default, there is no tab selected in the Config panel
noTabSelectedInConfigPanel

unsortedhashsets:
I copy my view on that problem from #1433:

ConfigTab use `useState` with the default value `configTabIndex`, but it works only for the initial Tab open, with clicking on another element it use `useEffect` without change of `configTabIndex` but it can be different for example here Choice config tab is extension and it has index 0 so without useEffect if you click on some other  step you will see a tab with index 0 (empty) and also in reverse after some general step you open choice with configtab index 2 (empty)

With that update switch between steps (with open details/config  VisualizationStepViews) always open the config tab as default

fixes #1433

### DIFF
--- a/src/components/VisualizationStepViews.tsx
+++ b/src/components/VisualizationStepViews.tsx
@@ -77,6 +77,7 @@ const VisualizationStepViews = ({
   };
 
   useEffect(() => {
+    setActiveTabKey(configTabIndex);
     let tempSchemaObject: { [label: string]: { type: string; value?: any; description?: string } } =
       {};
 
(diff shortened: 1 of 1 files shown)


kahboom (approved):
also fixes https://github.com/KaotoIO/kaoto-ui/issues/1280

PR #1452, opened 2023-03-10
Hide Add Branch button for steps with custom extensions

### LINKED ISSUE #1381: Adding new branch using the "Add a branch" popover not reflected in the code editor
### Describe the Bug

When user adds new branch for the Choice step using the "Add a branch" popover -   it adds a otherwise step:
```
otherwise:
      steps: []
```
This "otherwise" step can be added multiple times. Also the new branch is not reflected in the code editor.

1. Add a branch to choice step using the "Add a branch" popover
2. It's not visible in the editor

### Screenshots or Videos

Peek 2023-03-03 14-43

### Platform

- OS: [Linux]
- Browser: [Firefox]
- Version: [latest]

lordrip:
### Description
Taking as an example the `choice` EIP, currently there's no mechanism in place to determine whether a newly created branch it's from a `when` or `otherwise` type of branch hence the need for the custom step extension.

For said steps, displaying the `Add branch` button to the users might be confusing as adding empty branches leads to creating only one type of branch, also, if steps are added on said branches, those steps are lost upon syncing with the backend :cry: .

### Changes
* Extract a section of the `VisualizationStep` component into a dedicated `AppendStep` component to ease testing and as an initial step to remove not necessary logic from the `VisualizationStep` component
* Evaluate whether a given step has a Custom Step View in order to disable the `Add branch` button and display a tooltip explaining the reason to the user
image
* In case a step doesn't have a Custom Step View, we allow creating branches from the Mini Catalog :tm: 
image

Fixes #1381

### DIFF
--- a/src/components/AppendStepButton.tsx
+++ b/src/components/AppendStepButton.tsx
@@ -35,7 +35,7 @@ export const AppendStepButton: FunctionComponent<IAddStepButton> = ({
 
   useEffect(() => {
     if (hasCustomStepExtension) {
-      setDisableBranchesTabMsg(`The "${step.name}" step has a Custom Step Extension, please click on the step to configure it`);
+      setDisableBranchesTabMsg("Please click on the step to configure branches for it.");
       return;
     }
 
--- a/src/components/AppendStepButton.tsx
+++ b/src/components/AppendStepButton.tsx
@@ -1,9 +1,9 @@
-import { ValidationService } from '@kaoto/services';
-import { useSettingsStore } from '@kaoto/store';
+import { StepsService, ValidationService } from '@kaoto/services';
+import { useIntegrationJsonStore, useSettingsStore } from '@kaoto/store';
 import { IStepProps } from '@kaoto/types';
 import { Popover, Tooltip } from '@patternfly/react-core';
 import { PlusIcon } from '@patternfly/react-icons';
-import { FunctionComponent } from 'react';
+import { FunctionComponent, useEffect, useState } from 'react';
 import { BranchBuilder } from './BranchBuilder';
 import { MiniCatalog } from './MiniCatalog';
 
@@ -16,7 +16,7 @@ interface IAddStepButton {
   supportsBranching: boolean;
 }
 
-export const AddStepButton: FunctionComponent<IAddStepButton> = ({
+export const AppendStepButton: FunctionComponent<IAddStepButton> = ({
   handleAddBranch,
   handleSelectStep,
   step,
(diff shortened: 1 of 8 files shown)


src/components/VisualizationStep.tsx
```diff
@@ -207,53 +208,16 @@ const VisualizationStep = ({ data }: NodeProps<IVizStepNodeData>) => {
...
+          {VisualizationService.showAppendStepButton(data, endStep)
+            && <AppendStepButton
+              handleAddBranch={handleAddBranch}
+              handleSelectStep={onMiniCatalogClickAppend}
+              step={data.step}
+              showBranchesTab={showBranchesTab}
+              showStepsTab={showStepsTab}
+              supportsBranching={supportsBranching}
+            />
+          }
```
lordrip, line 220: This is the initial step, IMHO, the next candidate to move into its own component should be the `Add step` placeholder.

kahboom: I'm very happy with this, as I've been meaning to get some of these components moved into smaller ones for a while! FYI, you can also just write `showStepsTab` instead of `showStepsTab={showStepsTab}`. I don't mind either way, but just wanted to share in case you didn't already know!

igarashitm (dismissed):
LGTM :+1:

src/components/AppendStepButton.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const [hasCustomStepExtension, setHasCustomStepExtension] = useState(StepsService.hasCustomStepExtension(step, views));
+  const [disableBranchesTabMsg, setDisableBranchesTabMsg] = useState('');
+
+  useEffect(() => {
+    setHasCustomStepExtension(StepsService.hasCustomStepExtension(step, views));
+  }, [step, views])
+
+  useEffect(() => {
+    if (hasCustomStepExtension) {
+      setDisableBranchesTabMsg(`The "${step.name}" step has a Custom Step Extension, please click on the step to configure it`);
```
kahboom, line 38: I do wonder if the language should be simpler for the end user who doesn't know what a step extension is (my guess is that the person who sets it up would be the developer, and the one to use it would be the non-dev).

For example, maybe just saying `"Please click on the step to configure branches for it."` is enough? However, the user might wonder why this only applies to certain steps.. so I'm not 100% sure. 🤔  @igarashitm @mmelko

igarashitm: Ah good point, the word `Step Extension` would be a tech term for Kaoto developers. Can we say `click on the step to configure branches`? Not sure if it's always branches

src/components/VisualizationStep.tsx
lordrip: > but just wanted to share in case you didn't already know!

@kahboom thanks for sharing, I didn't know, but just to clarify, wouldn't that force to `showStepsTab` property to be true always?

src/components/AppendStepButton.tsx
lordrip: @igarashitm we shouldn't forget the magic word :stuck_out_tongue_closed_eyes:
 "please"

lordrip: I think that you're right, I'll update the pull request to use your example :bow:

PR #1474, opened 2023-03-14
fix(1379): improve layouting

### LINKED ISSUE #1146: UX: Increase spacing for steps with branches

If a step has multiple branches, the gap between the "root step" and its branches does not have much space, and the label looks crowded:

Screen Shot 2023-01-19 at 10 26 23 pm copy

or, vertically:

Screen Shot 2023-01-23 at 5 02 28 pm

An example with single step branches:

Screenshot from 2023-02-13 12-49-25

To fix this, we probably need to adjust some ELK configuration setting [here](https://github.com/KaotoIO/kaoto-ui/blob/40352b0151fb81f5e87aeb1cef627bd0b1cf2083/src/services/visualizationService.ts#L391) in `getLayoutedElements`.

### LINKED ISSUE #1374: Branching messed up with 3 branches with one having 2 steps
### Describe the Bug

Branching messed up with 3 branches with one having 2 steps

for instance using this route:

```
- from:
    uri: timer:null
    steps:
    - choice:
        when:
        - simple: path1
          steps:
          - to:
              uri: https:null
        - simple: path2
          steps:
          - log: {}
        - simple: path3
          steps:
          - to:
              uri: avro:null:null:null:null
          - log: {}
        otherwise:
          steps:
          - log: {}
    - log: {}
```

1. add a step
2. add a choice
3. add 3 branches
4. add 2 steps on one of the branches

### Screenshots or Videos

image

messedUpbranch

### Platform

OperateFirst

### LINKED ISSUE #1379: Delete branch button sometimes hidden by another branch step
### Describe the Bug

The delete branch element is hidden behind another branch step element. See screenshot

1. create several branches - ideally with multiple steps
2. Add a branch with single step 

### Screenshots or Videos

Screenshot from 2023-03-03 14-24-00

### Platform

- OS: [Linux]
- Browser: [Firefox]
- Version: [latest]

kahboom:
This PR fixes an issue where the "delete branch" button is covered by a step label, which is caused by the layouting. It fixes #1379, fixes #1374, and resolves #1146.

## Changes
- Change DAG layouting engine
- Update default layout strings
- Improve separation between rankings of nodes
- Increase weight of edge connecting first steps in a branch (to shift unbalanced branch nodes to be left-aligned)
- Adds a reference to the elkjs branch in the services `README` for reference

There is still some room for improvement:
- Add proper typings for the library
- It would be nice if the graph were symmetrical (for uneven branches, for example), so that edges are the same regardless of the number of nodes in a branch (see "After" of first example below)
- Maybe spacing should also be increased at certain places for vertical mode

## Screenshots

Before (delete branch button)

Screen Shot 2023-03-14 at 11 32 11 am copy

After

Screen Shot 2023-03-13 at 7 59 00 pm

Before (unbalanced branches)

Screen Shot 2023-03-14 at 11 34 40 am

After

Screen Shot 2023-03-14 at 11 36 05 am

Before (spacing of labels for nested branches in vertical mode)

Screen Shot 2023-03-14 at 11 38 18 am

After

Screen Shot 2023-03-14 at 11 37 48 am

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "@patternfly/react-table": "4.112.39",
     "@rhoas/app-services-ui-shared": "^0.16.6",
     "ajv": "^8.12.0",
+    "dagre": "^0.8.5",
     "elkjs": "^0.8.2",
     "immer": "^9.0.19",
     "lodash.isequal": "^4.5.0",
--- a/package.json
+++ b/package.json
@@ -55,7 +55,6 @@
     "@rhoas/app-services-ui-shared": "^0.16.6",
     "ajv": "^8.12.0",
     "dagre": "^0.8.5",
-    "elkjs": "^0.8.2",
     "immer": "^9.0.19",
     "lodash.isequal": "^4.5.0",
     "monaco-editor": "^0.36.1",
--- a/src/services/visualizationService.ts
+++ b/src/services/visualizationService.ts
@@ -14,7 +14,9 @@ import {
   IVizStepPropsEdge,
 } from '@kaoto/types';
 import { getRandomArbitraryNumber, truncateString } from '@kaoto/utils';
-import ELK, { ElkExtendedEdge, ElkNode } from 'elkjs';
+// import ELK, { ElkExtendedEdge, ElkNode } from 'elkjs';
+// @ts-ignore
+import dagre from 'dagre';
 import { MarkerType, Position } from 'reactflow';
 
 /**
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -54,6 +54,7 @@
     "@patternfly/react-table": "4.112.39",
     "@rhoas/app-services-ui-shared": "^0.16.6",
     "ajv": "^8.12.0",
+    "dagre": "^0.8.5",
```
apupier, line 57: dagre added?

src/services/visualizationService.ts
```diff
@@ -14,7 +14,8 @@ import {
   IVizStepPropsEdge,
 } from '@kaoto/types';
 import { getRandomArbitraryNumber, truncateString } from '@kaoto/utils';
-import ELK, { ElkExtendedEdge, ElkNode } from 'elkjs';
+// @ts-ignore
```
apupier, line 17: why do we need a ts-ignore?

apupier (commented):
- Is Elk still used in some places? If not, it should be removed from dependencies.
- Still some little glitches but it is a lot better:
image
```
- from:
    uri: timer:null
    steps:
    - choice:
        when:
        - steps:
          - to:
              uri: activemq:queue:null
        - steps:
          - to:
              uri: amqp:queue:null
          - to:
              uri: atmos:null:null
          - to:
              uri: atmos:null:null
        - steps:
          - to:
              uri: as2:null:null
        - steps: []
    - log: {}
```

lhein:
hmm...the before and after pictures have still glitches both. Can we eliminate that? Otherwise I am wondering if it is worth the risk of changing the layout engine.

igarashitm (dismissed):
LGTM :+1: I've been eager to get the 1st branch step aligned so that the conditions shown sorted, and I couldn't make it work with ELK (maybe with next 0.9 release... but we can't wait)

src/services/visualizationService.ts
kahboom: because there are no typings available for Dagre, we would need to provide our own or see if this exists already

src/store/visualizationStore.tsx
```diff
@@ -43,7 +43,7 @@ export const useVisualizationStore = create<RFState>((set, get) => ({
       nodes: [...state.nodes.filter((_n, idx) => idx !== nodeIndex)],
     })),
   hoverStepUuid: '',
-  layout: 'RIGHT', // e.g. RIGHT, DOWN
+  layout: 'LR', // e.g. LR, TB
```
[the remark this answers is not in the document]
kahboom: each layouting engine using a different string for layouting, so `LR` (left-right) and `TB` (top-bottom) is specific to Dagre and the `RIGHT` `DOWN` is specific to ELK.

igarashitm (approved):
:+1:

apupier (approved):

PR #1477, opened 2023-03-14
chore: StepsService cleanup

lordrip:
### Description
As a byproduct of another investigation, there are some small cosmetic improvements over the StepsService

### Changes
* Rename the `saveConfig` parameter to updateStepParams to (hopefully) make it easier to understand where this parameter goes
* Replace `.map()` call with a `.forEach` to avoid creating a new array
* Type the `buildStepSchemaAndModel` method parameters

### DIFF
--- a/src/services/stepsService.ts
+++ b/src/services/stepsService.ts
@@ -22,6 +22,7 @@ import {
   IStepProps,
   IStepPropsBranch,
   IViewProps,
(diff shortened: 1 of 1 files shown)


igarashitm (approved):

kahboom (approved):
thanks @lordrip ! I like the improvements

PR #1485, opened 2023-03-15
fix(viz): alignment append/prepend step 

### LINKED ISSUE #1184: Branching: Append button should be below step in vertical mode
### Describe the Bug

If you are in vertical mode and have an integration with branches and no sink step, you see an append button (as you should), but the append button shows next to it, not below it like the insert step button does.

1. Go to Settings
2. Change DSL to `Integration`, press save
3. Open the code editor
4. Copy and paste YAML below, press save
5. Switch to vertical mode by pressing the button on the far lower right hand corner of the visualization
6. See the plus button is to the right of the step for append, but is correct for insert

### Screenshots or Videos

Screen Shot 2023-01-27 at 1 05 48 pm

### Platform

- Browser: Chrome
- Version: 0.6.2-dev

kahboom:
This PR fixes an issue where, in vertical mode, the append/prepend step buttons were showing to the left and right of a step, instead of above and below, respectively. fixes https://github.com/KaotoIO/kaoto-ui/issues/1184

## Changes
- Move append step button to below steps in vertical mode
- Mode prepend step button to above steps in vertical mode
- Bonus bug: Position tooltips to the left or right instead of top/bottom in vertical mode
- Improve handle aesthetics to match the rest of the step node

## Screenshots

Prepend and append step button before:
Screen Shot 2023-03-15 at 12 27 05 pm copy

Prepend and append step after:
Screen Shot 2023-03-15 at 2 56 57 pm

Tooltips in vertical mode before:
Screen Shot 2023-03-15 at 3 58 14 pm

Tooltips in vertical mode after:
Screen Shot 2023-03-15 at 3 06 46 pm

Screen Shot 2023-03-15 at 3 09 32 pm

Handle before (top) and after (bottom):
Screen Shot 2023-03-15 at 1 18 42 pm

Handle after:

Screen Shot 2023-03-15 at 2 56 49 pm

Screen Shot 2023-03-15 at 3 22 20 pm

### DIFF
--- a/src/components/AppendStepButton.tsx
+++ b/src/components/AppendStepButton.tsx
@@ -1,15 +1,16 @@
+import { BranchBuilder } from './BranchBuilder';
+import { MiniCatalog } from './MiniCatalog';
 import { StepsService, ValidationService } from '@kaoto/services';
 import { useIntegrationJsonStore, useSettingsStore } from '@kaoto/store';
 import { IStepProps } from '@kaoto/types';
 import { Popover, Tooltip } from '@patternfly/react-core';
 import { PlusIcon } from '@patternfly/react-icons';
 import { FunctionComponent, useEffect, useState } from 'react';
-import { BranchBuilder } from './BranchBuilder';
-import { MiniCatalog } from './MiniCatalog';
 
 interface IAddStepButton {
   handleAddBranch: () => void;
   handleSelectStep: (selectedStep: IStepProps) => void;
+  layout: string;
   step: IStepProps;
   showBranchesTab: boolean;
   showStepsTab: boolean;
@@ -19,6 +20,7 @@ interface IAddStepButton {
 export const AppendStepButton: FunctionComponent<IAddStepButton> = ({
   handleAddBranch,
   handleSelectStep,
+  layout,
   step,
   showBranchesTab,
   showStepsTab,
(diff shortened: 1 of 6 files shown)


igarashitm (approved):

src/components/Visualization.css
```diff
@@ -22,6 +22,13 @@
     height: calc(100vh - 77px);
 }
 
+.stepHandle {
+    width: 8px !important;
```
kahboom, line 26: @lordrip I tried to find a way around having to use `!important` here but couldn't figure it out. If you have any ideas though let me know! :)

src/components/AppendStepButton.tsx
```diff
@@ -77,14 +83,17 @@ export const AppendStepButton: FunctionComponent<IAddStepButton> = ({
     >
       <Tooltip
         content={ValidationService.getPlusButtonTooltipMsg(showBranchesTab, showStepsTab)}
+        position={layout === 'LR' ? 'top' : 'right'}
       >
         <button
-          className="stepNode__Add plusButton nodrag"
+          className={`${
+            layout === 'LR' ? 'stepNode__Add' : 'stepNode__Add--vertical'
+          } plusButton nodrag`}
```
lordrip, line 91: [minor]: I think that in this case, both evaluations will run every time since we're defining those objects here. So what are your thoughts about having something like:

```typescript
// This can be somewhere else or used with useCallback() instead.
const calcClassName = (layout: string): string => {
    const className =  'LR' ? 'stepNode__Add' : 'stepNode__Add--vertical';
    return className + ' plusButton nodrag';
};

const [tooltipPosition, setTooltipPosition] = useState(layout === 'LR' ? 'top' : 'right');
const [buttonClassName, setButtonClassName] = useState(calcClassName(layout));

useEffect(() => {
    setTooltipPosition(layout === 'LR' ? 'top' : 'right');
    setButtonClassName(calcClassName(layout));
}, [layout]);
```

and then lastly just bind the variables in the template part

```jsx
<Tooltip
  content={ValidationService.getPlusButtonTooltipMsg(showBranchesTab, showStepsTab)}
  position={tooltipPosition}
  >
        <button
          className={buttonClassName}
          ...
        />
</Tooltip>
```

src/components/Visualization.css
lordrip: mmm I see :thinking:, I think that it's similar to what @igarashitm had the other day, in relation to the style specificity.

One way to remove the `!important` bit could be the following:
```css
.stepNode .stepHandle.react-flow__handle {
    width: 8px;
    height: 8px;
    background: var(--pf-global--BorderColor--200);
    border-radius: 100%;
}
```
it's the same style as before but now I tried to provide more specificity, if I'm not mistaken, we're saying now:

> _All children from nodes containing the `.stepNode` class AND that also have the `stepHandle` and `react-flow__handle` classes together_

That being said, I was curious why the React flow class won instead of ours, and it seems that it's related to the definition order, probably the react flow styles are added below ours, hence getting the winning hand for the colliding specificity, but this is just a theory :sweat_smile: 

I found [this link](https://www.w3.org/TR/CSS2/cascade.html#cascading-order) from an older draft very interesting in regards to what gets applied when. :smiley:

lordrip (approved):
LGTM

PR #1487, opened 2023-03-15
fix: Use condition for branch label instead of identifier

### LINKED ISSUE #1486: Use condition for branch label instead of identifier
### Describe the Bug

cf. https://github.com/KaotoIO/kaoto-backend/issues/513
cf. https://github.com/KaotoIO/step-extension-repository/pull/346

In order to allow backend to distinguish `when` with empty condition and `otherwise`, choice step extension will use `when-0`, `when-1` ... for the branch `identifier`.  The side effect is that the branch label shown in the UI canvas showing `identifier` as a label. We'd want to show `condition` instead. If the condition is `null`, then it alternatively can show `identifier` (which helps for otherwise branch).

1. Load the choice step extension with this change https://github.com/KaotoIO/step-extension-repository/pull/347
2. Add `choice` step and add multiple `when`
3. See the branch label, it will be `when-0`, `when-1`, ...

### Screenshots or Videos

_No response_

### Platform

- Version: [e.g. 0.30.0]

igarashitm:
Fixes: #1486

### DIFF
--- a/src/services/visualizationService.ts
+++ b/src/services/visualizationService.ts
@@ -361,7 +361,7 @@ export class VisualizationService {
         step.branches.forEach((branch) => {
           stepNodes = stepNodes.concat(
             VisualizationService.buildNodesFromSteps(branch.steps, layout, props, {
-              branchIdentifier: branch.identifier,
+              branchIdentifier: branch.condition !== null ? branch.condition : branch.identifier,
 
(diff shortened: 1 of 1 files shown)


Delawen (approved):

kahboom:
>I believe `branchIdentifier` here is only used for the branch label, but I'd love to hear from you!

hey @igarashitm , yep that's correct

PR #1488, opened 2023-03-15
fix: Hide parameters that are not supported in the default config panel

### LINKED ISSUE #1444: Hide parameters that are not supported in the default config panel
### Describe the Bug

If you open up the config panel for a step you will often find bold labels which (from a user pov) look like headlines or section headers but instead these are labels of parameters which cannot be changed in the generic step extension. This is a confusing experience for users.

I suggest to not show these parameters at all or at least not making them appear in bold.

1. drag a component on the canvas (for instance git)
2. click on the component
3. look at the config panel

### Screenshots or Videos

grafik

### Platform

- Version: [e.g. 0.30.0]

igarashitm:
Fixes: #1444

Screenshot from 2023-03-15 15-36-48

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,20 @@
 import './App.css';
+import { useTemporalIntegrationJsonStore } from './store';
+import { bindUndoRedo } from './utils';
 import { AlertProvider, AppLayout, MASLoading } from '@kaoto/layout';
 import { AppRoutes } from '@kaoto/routes';
 import '@patternfly/patternfly/patternfly-theme-dark.css';
 import '@patternfly/patternfly/patternfly.css';
 import '@patternfly/patternfly/utilities/Accessibility/accessibility.css';
+import '@patternfly/patternfly/utilities/BackgroundColor/BackgroundColor.css';
 import '@patternfly/patternfly/utilities/Display/display.css';
 import '@patternfly/patternfly/utilities/Flex/flex.css';
 import '@patternfly/patternfly/utilities/Sizing/sizing.css';
 import '@patternfly/patternfly/utilities/Spacing/spacing.css';
-import 'reactflow/dist/style.css';
+import '@patternfly/patternfly/utilities/Text/text.css';
 import { Suspense } from 'react';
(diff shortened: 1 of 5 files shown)


lhein:
Its not really hiding but it improves the situation already a lot.

igarashitm:
We can also hide, it's much easier. But I prefer this to indicate it's still configurable from source code editor.

apupier (dismissed):

lordrip (dismissed):
LGTM, my only question would be about the text sizing. I'm wondering if would be better to increase the text for the text description and keep small the text about configuring it in the code editor, but just cosmetic stuff :nail_care:

igarashitm:
@lordrip 
So those are patternfly `<Popover>` default styles, where `Please use source code editor...` is in `headerContent`, and the parameter explanation is in `bodyContent`.
https://github.com/KaotoIO/kaoto-ui/pull/1488/files#diff-619ce600eab9984e34fdfab6c8992ba2e4f1d2d4be10ffc8d51854882fdca422R20

If we change the style for the parameter explanation font, we'd be better to do for all parameters but not only the disabled parameter. Do you think `Please use source code editor...` thing is better in `bodyContent`?

kahboom (dismissed):
Looks good to me, only thing I would recommend is to add some spacing between the fields for aesthetic reasons, but that's not specific to this PR. We can look at that for a tiny enhancement later on. 👍

igarashitm:
@kahboom that's I was also wondering, let me do that in this PR :+1:

igarashitm:
OK, that was not that easy. Maybe need an uniforms custom theme? or just use one of uniforms OOTB theme, as the current style looks like `unstyled`.  Also I found that `uniforms` and `uniforms-patternfly` are mixed ATM
https://github.com/KaotoIO/kaoto-ui/pull/1488/files#diff-2f2dde39f53ead9cb0ebae0fb587026d144104054d21ed5f0daa046e8d5e0fe6R4
I quickly tried to use `uniforms-patternfly` version of `<AutoForm>` but got a weird error. Maybe better to investigate in a separate issue.

kahboom (approved):
👍

PR #1505, opened 2023-03-17
chore: Add more space for vertical mode

lordrip:
### Description
Add more vertical spacing for vertical mode

|**Before** |**After** |
| --- | --- |
|image | image |

This is particularly relevant for the merging part

|**Before** |**After** |
| --- | --- |
| image | image |


kahboom:
They steps in the main integration are quite separated but we can always add some more elaborate condition later on for that, this is a big improvement

kahboom (approved):

PR #1514, opened 2023-03-20
chore: Add hooks missing tests

lordrip:
#### Adding hooks missing tests


src/hooks/position.hook.test.tsx
```diff
@@ -0,0 +1,67 @@
+import { TooltipPosition } from '@patternfly/react-core';
+import { render } from '@testing-library/react';
+import { act } from 'react-dom/test-utils';
+import { Position } from 'reactflow';
+import { useVisualizationStore } from '@kaoto/store';
+import { usePosition } from './position.hook';
```
kahboom, line 6: I know this is a late thought but I do wonder if we could have a better name than `usePosition` or `position.hook`. At first glance I thought it was node position.

kahboom (approved):

Issue #1175, opened 2023-01-26
Settings: Theme choice doesn't persist correctly


kahboom:
### Describe the Bug

When you choose an alternate theme, like dark mode, it saves the setting to `localStorage`, but on refresh it will show light mode again. It's not until you go to the settings again that it refreshes the local state.

The solution would be to follow what we are doing for the code editor theme, which is persisted in `localStorage` and also stored in the `settingsStore`. See `SourceCodeEditor.tsx` and `AppearanceModal.tsx` components for reference.

### Steps to Reproduce the Bug or Issue

1. Go to Appearance settings
2. Toggle to Dark Mode for the UI theme
3. Close modal, see it's in dark mode
4. Refresh page, see it's in light mode again
5. Go back to Appearance settings, and it changes back to dark mode

### Screenshots or Videos

![Kapture 2023-01-26 at 20 43 13](https://user-images.githubusercontent.com/3844502/214946423-1ff840a0-83b0-44de-80eb-654560ecff33.gif)

### Platform

- OS: macOS
- Browser: Chrome
- Version: 0.6.1

tplevko:
Just updated my previous commit - found a way to make it work also in VSCode.

See https://github.com/KaotoIO/kaoto-ui/pull/1551

kahboom:
Don't worry @tplevko I am also to blame for not testing with VSCode locally and not including these in the e2e tests for 1.0, because I could not figure out yet a way to make this work (same thing for deployment testing). This would be great to explore next week if we have time, since we'll hopefully be on a feature freeze 🤞--even if it's not included until the following release.

apupier:
> how we could run tests with PR's - not only on Kaoto-ui, but also with the VSCode extension

This is a problem identified also with AtlasMap previously. Unfortunately, in several years, we never had time to start investigation on this kind of things.

PR #1531, opened 2023-03-21
Fix the UI theme switch - add to local storage

tplevko:
Fix for https://github.com/KaotoIO/kaoto-ui/issues/1175

Placed the theme configuration useEffect in the KaotoToolbar element - let me know, if this makes sense, or is there better choice of component? :)


kahboom (approved):
thank you @tplevko. I don't think it's too weird to put the `useEffect` there. if it causes problems later we can always move it.

Issue #1371, opened 2023-03-03
Config panel is not updated on Drag'n'drop from catalog to replace a step


apupier:
### Describe the Bug

Config panel is not updated on Drag'n'drop from catalog to replace a step

### Steps to Reproduce the Bug or Issue

1. add a step
2. select the step to have config panel opened
3. open step catalog
4. Drag'n'drop another step
5. the name of the step is replaced but the config panel is still on the previous step

### Screenshots or Videos

![notUpdatedpanelOnDnDFromStepCatalog](https://user-images.githubusercontent.com/1105127/222690788-a3a49c09-d1be-4c2b-92bc-1c5cb75bc7de.gif)

### Platform

OperateFirst

kahboom:
closed via https://github.com/KaotoIO/kaoto-ui/pull/1544

Issue #609, opened 2022-07-26
Persist Steps: On Refresh


kahboom:
Child issue for this epic: #607 

## Problem
On refresh, you lose all data. This includes settings and steps.

## Proposition
Store state of Visualization in `localStorage`.

ibek:
+1 for the proposal and addressing this problem. It happens that you misclick, go back in history, close a tab, whatever and you don't want to use all your work. Just restoring the latest YAML / session would be enough.

PR #1544, opened 2023-03-23
fix(stepDetail): close step detail panel if unable to locate step

### LINKED ISSUE #1475: Syncing the code loses the currently selected step if said step is nested
### Describe the Bug

Having selected a nested step, if the code is synced, this value gets lost. This might be caused by how the IDs are calculated [from this line](https://github.com/KaotoIO/kaoto-ui/blob/main/src/services/stepsService.ts#L470)

```typescript
if (branchSteps) step.UUID = `${step.name}-${idx}-${getRandomArbitraryNumber()}`;
```

1. Use the following `yaml` file as a starting point
```yaml
- from:
    uri: as2:null:null
    steps:
    - choice:
        when:
        - simple: 'This is a simple condition'
          steps: []
        otherwise:
          steps:
          - choice:
              when:
              - jq: 'This is a nested JQ condition'
                steps: []
```

2. Click on the nested `choice` step
3. Sync the code
4. Notice how the selected step gets lost and the view is not right

### Screenshots or Videos

[Screencast from 2023-03-14 14-16-48.webm](https://user-images.githubusercontent.com/16512618/225012806-0ab4ae3e-f3b1-45ac-bcf0-e9cd1ff43cd0.webm)

### Platform

- OS: [Linux]
- Browser: [Firefox]
- Version: [0.7.2-dev]

kahboom:
This PR fixes #1475 and #1371, where the updated step does not have the same `UUID` and then the open step detail view is incorrect. Now, if a step with the same `UUID` is not found, the step detail panel will close.

**NOTE:** This fixes the reported bugs, but the user experience is not ideal. We have discussed using the previously selected step's path to automatically select the new step in that position, or to try to search for the previous step, however, without persisted unique identifiers we are not able to consistently predict where the previous step will be and this could be perceived as a bug rather than a feature. #609 might be able to address this in the future, or we can consider some of the other workarounds discussed in #1475.

## Changes
- On change to `integrationJson`, close the step detail panel and reset the selected step if a step with a matching `UUID` is not found

Not sure what is happening with the code formatting again. 🤔


igarashitm (dismissed):

src/components/Visualization.tsx
```diff
@@ -60,6 +60,10 @@ const Visualization = () => {
     const step = stepsService.findStepWithUUID(selectedStep.UUID);
     if (step) {
       setSelectedStep(step);
+    } else {
+      setSelectedStep({ maxBranches: 0, minBranches: 0, name: '', type: '', UUID: '' });
+      visualizationStore.setSelectedStepUuid('');
+      if (isPanelExpanded) setIsPanelExpanded(false);
```
lordrip, line 66: Should we check whether it's closed or not? I think that this update will happen inside of an update batch, so I'm wondering if we can simply set it to `false` and call it a day :smile: 

```suggestion
      setIsPanelExpanded(false);
```

lordrip (dismissed):

igarashitm (approved):
:+1:

lordrip (approved):

PR #1545, opened 2023-03-23
fix(step-details): Add @patternfly standard spacing to form's field

### LINKED ISSUE #1495: Add spacing between parameters in default Config tab
### Describe the Bug

cf. https://github.com/KaotoIO/kaoto-ui/pull/1488

In default step details Config tab, parameters are too close and maybe better to add a bit spacing.

It might need an uniforms custom theme, or just use one of uniforms OOTB theme, as the current style looks like unstyled.
One thing to note is that in current code uniforms and uniforms-patternfly are mixed
https://github.com/KaotoIO/kaoto-ui/blob/4c8159294eabed57d5935ae28f0f3812c8566e2f/src/components/JsonSchemaConfigurator.tsx#L3
I quickly tried to use uniforms-patternfly version of `<AutoForm>` but got a weird error. 
And also check `@kie-tools/uniforms-patternfly` there as `uniforms-patternfly` is deprecated.
https://github.com/aerogear/uniforms-patternfly

1. Add activemq
2. Click added activemq step

### Screenshots or Videos

_No response_

### Platform

- Version: [e.g. 0.30.0]

lordrip:
### Description
At the moment all fields from the step's details are close to each other making them hard to read.

This is caused because there's a div between the form element and its children, meaning that the style cascade breaks.

### Changes
- Inherit the form styles to that intermediate div
- Remove the archived dependency `"uniforms-patternfly"` in favor of `"@kie-tools/uniforms-patternfly"`

| Before | After |
| --- | --- |
| Screenshot from 2023-03-23 18-38-49 | Screenshot from 2023-03-23 18-38-59 |

fixes: https://github.com/KaotoIO/kaoto-ui/issues/1495


igarashitm (dismissed):
:+1: from me

kahboom (dismissed):
jest may not be happy, but i am

lordrip:
> ah, maybe `jest-esm-transformer` needed? like this - https://github.com/atlasmap/atlasmap/blob/main/ui/packages/atlasmap/package.json#L123

Yes @igarashitm , it's because we're using the ESM version of `@kie-tools/uniforms-patternfly` but we're not providing any transformation for it in `jest`, as a workaround, I added a moduleNameMapper to use the `CJS` version during tests.

I'm not 100% that it's the best solution though, do you think that providing the other config would be better?

kahboom:
>Yes @igarashitm , it's because we're using the ESM version of @kie-tools/uniforms-patternfly but we're not providing any transformation for it in jest, as a workaround, I added a moduleNameMapper to use the CJS version during tests.
>
>I'm not 100% that it's the best solution though, do you think that providing the other config would be better?

I think `moduleNameMapper` is the most straightforward and recommended way to deal with these issues, but I'm not too familiar with using `jest-esm-transformer` to know for sure. If it's working then that's the important thing. 😄

kahboom (approved):

PR #1550, opened 2023-03-24
fix(vscode): Sort Step's parameters in VS Code

### LINKED ISSUE #1301: Mandatory fields in the step's config panel

Some fields are mandatory in the configuration panel of a step. These are marked with a * but atm they are not easy to spot if the step has a lot of configuration options (for instance activemq or kafka components). 

How about showing them in a separate tab or alternatively at the top of the configuration page (maybe grouped as 'Mandatory')?

relates also to KaotoIO/kaoto-backend#474

lordrip:
### Description
Originally, the step's parameters order was based on the premise that the properties order is kept when going from `Array > Object` in modern browsers, but also that using the ApiService was the only place in which a step detail was fetched.

Both assumptions were not completely correct, so now the sort strategy it's based on the AutoForm library which allows specifying a fixed order for the automated generated fields.

fixes https://github.com/KaotoIO/kaoto-ui/issues/1301 (again)


apupier (approved):
Now working fine in standalone using the sync button and in VS Code:
image
correctOrderWithSyncCode

Issue #1099, opened 2023-01-13
Re-enable e2e tests


kahboom:
### Please describe the task that needs to be done

Now that the main issues with the code editor have been resolved, we should be able to add back e2e tests.

- [x] Create common happy paths
- [x] Update existing e2e tests
- [x] Add any missing e2e tests
- [x] Update relevant API keys where necessary
- [x] Ensure e2e tests are run on each PR

Tests should only be run against the `main` branch, not for forks. Relates to #1100.

kahboom:
@tplevko no worries I will take a look, thank you for all of your help with this!

tplevko:
@kahboom closing this as the e2e tests are already reenabled on PR's. Thanks for helping out with this.

PR #1560, opened 2023-03-27
Refactor e2e tests 4/5/6/8, fix autoScrolling issue

unsortedhashsets:
Hi!

With that PR I want to update the new part of E2E tests (smaller + stable + with common commands)

Also fixed the strange issue in the cypress dashboard with the lowest (hidden!) frame resizing (issue was faced only in cypress, I did not found any related issue) -> `scrollBehavior: false`

Before:
Screenshot from 2023-03-27 13-22-45

After:
Screenshot from 2023-03-27 13-23-49 

Related: 
- [Issue-1099](https://github.com/KaotoIO/kaoto-ui/issues/1099) (can be related to API keys part and also general stabilization)

### DIFF
--- a/cypress/support/kaoto-ui-commands/editor.js
+++ b/cypress/support/kaoto-ui-commands/editor.js
@@ -1,8 +1,11 @@
 import 'cypress-file-upload';
 
-Cypress.Commands.add('uploadInitialState', (fixture) => {
-    cy.get('[data-testid="toolbar-show-code-btn"]').click();
-    cy.wait('@getIntegration');
+Cypress.Commands.add('uploadFixture', (fixture, initial) => {
+    initial = initial ?? true;
+    if (initial) {
+        cy.get('[data-testid="toolbar-show-code-btn"]').click();
+        cy.wait('@getIntegration');
+    }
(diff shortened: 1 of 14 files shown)


cypress/support/kaoto-ui-commands/editor.js
```diff
@@ -35,6 +38,13 @@ Cypress.Commands.add('editorDeleteLine', (line, repeatCount) => {
         );
 });
 
+Cypress.Commands.add('editorClickUndoXTimes', (times) => {
+    times = times ?? 1;
+    Cypress._.times(times, () => {
```
kahboom, line 43: I wonder if we are using lodash anywhere else (we probably are), though it would be good to try to remove it if possible. This would be a good alternative for this method in the future: https://youmightnotneed.com/lodash/#times

kahboom (approved):

cypress/support/kaoto-ui-commands/editor.js
unsortedhashsets: Thanks, will fix it in the last refactor PR

PR #1578, opened 2023-03-28
fix(StepsService): Generate steady IDs for steps and branches

### LINKED ISSUE #1553: Config panel is closing when editing text fields of a step inside a branch in VS Code
### Describe the Bug

* Create route with a choice with 2 branches
* add steps on each of the branch
* select one
* type something in a text field
--> the config panel is closing

* Create route with a choice with 2 branches
* add steps on each of the branch
* select one
* type something in a text field
--> the config panel is closing

Note that sometimes it is not happening but most of the time it is

### Screenshots or Videos

closingConfigPanel

### Platform

VS Code using main branch of Kaoto UI

lordrip:
### Context
At the moment, when a step is no longer available because it was removed, or its id changed, we close the step's detail panel since is no longer editing the right step.

This causes an issue with `vscode` since there's a circular dependency between the integration JSON and the code editor. Long story short, whenever there's an update of the integration JSON, the steps or branch IDs change, forcing a new sync cycle from the `kogito` integration side.

A partial fix to this issue is to ensure that the IDs are consistent between syncs, so this way [the comparison inside of the KogitoEditorIntegrationProvider.tsx](https://github.com/KaotoIO/kaoto-ui/blob/107550559f099d391d94cc94aab7a58203ca8548/src/kogito-integration/KogitoEditorIntegrationProvider.tsx#L95) works.

```typescript
  // Update KaotoEditor content after an integrationJson change (the user interacted with the Kaoto UI).
  useCancelableEffect(
    useCallback(
      ({ canceled }) => {
        if (!integrationJson || isEqual(previousJson.current, integrationJson)) return;
...
```

### Changes
- Remove the usage of `getRandomArbitraryNumber()` to calculate IDs
- Generate IDs based on:
    - Step name
    - Step position
    - Branch position

Considering the following YAML file
```yaml
- from:
    uri: activemq:queue:null
    steps:
    - choice:
        when:
        - simple: '{{header}}'
          steps:
          - to:
              uri: sql:null
        - simple: '{{body}}'
          steps:
          - to:
              uri: mongodb:null
        otherwise:
          steps:
          - to:
              uri: sql-stored:null
```

the IDs would be:

```
- activemq: activemq-0
- choice: choice-1
    - sql: choice-1|branch-0|sql-0
    - mongodb: choice-1|branch-1|mongodb-0
    - sql-stored: choice-1|branch-2|sqlstored-0
```

fix: https://github.com/KaotoIO/kaoto-ui/issues/1553

related to: https://github.com/KaotoIO/kaoto-ui/issues/1475#issuecomment-1479889304


apupier (approved):
Not looked at the code.

From a functional point of view, in VS Code, the panel is no more closing. Need to combine testing with the other PR otherwise data is lost and it become crazy https://github.com/KaotoIO/kaoto-ui/pull/1580

apupier (approved):
From functional point of view, I tried with VS Code and it is working better. No more able to reproduce the data loss.

EDIT: argh this was for the other PR https://github.com/KaotoIO/kaoto-ui/pull/1580

kahboom (approved):

Issue #1258, opened 2023-02-14
Add mocking for testing


kahboom:
### Please describe the task that needs to be done

We need to add support for mocking both the API as well as the stores, which can be used in E2E, component, and unit tests.

## Tasks
- [ ] Add API mocking support
- [ ] Add store mocking support

## Other considerations
- What other services or libraries do we need to mock? Monaco?
- How can we mock a cluster or similar, to test deployments? Is this possible or even feasible?

Issue #1100, opened 2023-01-13
Re-enable Storybook


kahboom:
### Please describe the task that needs to be done

We should look to re-enable Storybook to continue to ensure that our code base has the separation of concerns for visual components.

- [x] Re-add Storybook
- [x] Fix Storybook (again) after class implementation, unless #1233 gets resolved beforehand
- [x] Add MSW to mock network requests where necessary (see #1258, hopefully we can use this mocking for both types of tests)
- [x] Add stories for components
- [x] Add documentation for components, where possible
- [x] Update relevant API keys where necessary
- [x] Ensure Storybook stories are run on each PR
- ~~Test the stories, whether through React Testing Library or Cypress component testing~~ TBD, might add too much overhead

Tests should only be run against the `main` branch and primary repo, not for forks. This could be good to do once #1183 and other refactoring of stateful components is done.

kahboom:
@igarashitm - Originally I wanted to wait to have this once we refactored the UI components into stateless components, but I think it would be good to already start building them again and creating basic stories and docs for the components like this: https://storybook.js.org/addons/@storybook/addon-docs

What do you think? Or wait until the refactor (depending what kind of refactoring you decide to do)? I wouldn't start on this right away but at the end of next week if I finished visual branching and e2e tests in time.

igarashitm:
definitely +1, storybook is worth paying the cost to maintain, useful for design discussion as well as for detecting visual regression earlier IMO.

igarashitm:
In theory internal refactoring shouldn't conflict with storybook work - maybe little bit in mock data, but should be manageable.

PR #1579, opened 2023-03-28
chore: re-enable storybook, add support for mocking

kahboom:
This PR adds msw, sets up mocking for Storybook, and sets up an example call. It only mocks one endpoint as an example so that we can continue to work on this once 1.0 is released. It also re-enables Storybook, fixing an issue with private classes. Relates to #1100 and #1258 , but doesn't close them.

Please let me know if you'd rather calls be mocked using JS/TS instead of JSON. I chose JSON since it's what the API returns and is the fastest performance-wise, but I don't have a strong opinion about this.

Storybook Catalog story:

Screen Shot 2023-03-28 at 2 25 30 pm


igarashitm (approved):
:tada:

PR #1537, opened 2023-03-22
fix(step-details): Debounce saveConfig callback

lordrip:
### Context
At the moment of writing, whenever a user writes a character, the saveConfig callback is executed and 2 HTTP calls are triggered to update the YAML integration file.

This becomes problematic for bigger integrations or for long strings because triggers unnecessary server calls.

### Changes
This commit debounces that call by 500ms to leave time for the user to finish writing before updating the YAML file.

[Screencast from 2023-03-22 16-19-58.webm](https://user-images.githubusercontent.com/16512618/226952311-1a3a964e-9aba-467a-9182-9adc1f4cbc04.webm)

fixes https://github.com/KaotoIO/kaoto-ui/issues/1339

PR #1580, opened 2023-03-29
fix(step-details): Step's details panel is referencing an old value

### LINKED ISSUE #1561: Sometimes the provided data in config panel field are ignored
### Describe the Bug

Sometimes the provided dat avalues are ignored. i wasn't able to determine steps to reproduce but it happens quite regularly (like very 4 or 5 times I try to modify a different step)

I found only this stack in log:
```
stepsService.ts:539 Uncaught TypeError: Cannot set properties of undefined (setting 'value')
    at stepsService.ts:539:44
    at Array.forEach (<anonymous>)
    at StepsService.updateStepParameters (stepsService.ts:537:33)
    at Object.<anonymous> (Visualization.tsx:149:20)
    at invokeFunc (index.js:160:1)
    at trailingEdge (index.js:207:1)
    at timerExpired (index.js:195:1)
 ```

Not found

### Screenshots or Videos

dataVanished

### Platform

- OS: Fedora
- Browser:  VS Code and Firefox with OperateFirst
- Version: 0.7.5 and main branch on operateFirst

lordrip:
### Description
Currently, when opening the step’s details tab, it’s not referencing the most recent values but keeping the previous values from a previous step.

This generates unintentional changes to a different step than the user is intending to.

This situation was caused because of the addition of a `debounce` function with the intention to only save the [form’s values after the user’s input stopped](https://github.com/KaotoIO/kaoto-ui/pull/1537).

The problem was that the `debounce` function was causing to capture the form’s value from the previous render, hence saving the wrong values to the previous step.

To avoid this situation in the future, the saveConfig function now receives the original step and the new values.

### Changes
* Refactor `saveConfig` callback to receive the original step as a parameter
* Replace lodash' debounce function with `useDebounceCallback()`
* Increase the `debounce` timer from `500ms` to `1s`
* Update `onChangeModel` signature to `(configuration: Record<string, unknown>, isValid: boolean)` to reflect more accurately the intended parameters.
* Remove the `lodash.debounce` dependency since is no longer required

### Before
[Screencast from 2023-03-29 07-46-54.webm](https://user-images.githubusercontent.com/16512618/228438294-6f09f14c-278c-446a-a061-485b009ed068.webm)

### After
[Screencast from 2023-03-29 07-50-39.webm](https://user-images.githubusercontent.com/16512618/228438861-0af6eaf9-199f-43f2-8937-209f82bf620c.webm)

fix: https://github.com/KaotoIO/kaoto-ui/issues/1561


unsortedhashsets (dismissed):
LGTM, + tested with blocked E2E test

apupier (dismissed):
From functional point of view, I tried with VS Code and it is working better. No more able to reproduce the data loss.

kahboom (dismissed):
LGTM!

kahboom:
@lordrip only thing is that you have some dependency conflicts in `package.json` you'll have to resolve before merging

PR #1606, opened 2023-03-31
chore(test): add unit tests for all stores

kahboom:
This PR adds unit tests for all Zustand stores, since this was one of the areas with the lowest test coverage.

## Changes
- Add mock for Zustand (relates to #1258)
- Make parameters in stores easier to understand (for example, `replaceParentStep` instead of `replaceStep`, or `updateViews` instead of `setViews`)
- Add type to `updateIntegration` method in `visualizationStore`

**Tasks remaining:** Next PR will be focusing on other areas that are also lacking test coverage, such as `services` and `api`.

## Screenshots

Current code coverage for `store` (43.13%):

Screen Shot 2023-03-31 at 3 27 29 pm

It should be 100% after this PR is merged.

### DIFF
--- a/src/__mocks__/zustand.js
+++ b/src/__mocks__/zustand.js
@@ -1,0 +1,33 @@
+const { create: actualCreate, createStore } = jest.requireActual('zustand');
+const { act } = require('react-dom/test-utils');
+
+/**a variable to hold reset functions for all stores declared in the app */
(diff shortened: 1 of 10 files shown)


igarashitm (approved):
Great work :tada: will be much easier to add more tests from here, and prevent numerous possibilities of regression in the future :tada:

PR #1664, opened 2023-04-17
chore(viz): Adjust Node label clickable area

### LINKED ISSUE #1623: Make clickable <span> with step name, and disable clickable <div> + convert "ADD A STEP" step label into the span too

Hi, based on the issue [issues-1597)](https://github.com/KaotoIO/kaoto-ui/issues/1597)

During the investigation, I faced a few things that I would like to adjust a bit 

1) We have a clickable `<div>` container that is bigger than text content:
chrome-capture-2023-3-4

2) That container covered by non-clickable span if some step exists
chrome-capture-2023-3-4 (1)

I want to suggest:

1. Make container unclickable
2. Default value "ADD A STEP" add into the span as we have for other steps
3. Make span clickable

WDYT @apupier @kahboom?

Something like:
chrome-capture-2023-3-4 (2)

unsortedhashsets:
Hi,

Here is a very small visualization update - described here -> [issue-1623](https://github.com/KaotoIO/kaoto-ui/issues/1623)

Before:

1) We have a clickable `<div>` container that is bigger than text content:
chrome-capture-2023-3-4

2) That container covered by non-clickable span if some step exists
chrome-capture-2023-3-4 (1)

After:

chrome-capture-2023-3-4 (2)

If you have any suggestions feel free to write :smile:

### DIFF
--- a/src/components/Visualization.css
+++ b/src/components/Visualization.css
@@ -107,12 +107,14 @@ button[data-disable-branchestab="true"][data-disable-stepstab="true"]:hover {
     left: -75%;
     position: relative;
     width: 250%;
+    pointer-events: none;
 }
 
 .stepNode__Label span {
     background: var(--pf-global--BackgroundColor--150);
     border-radius: 15%;
     padding: 0.4em;
+    pointer-events: auto;
 }
 
 .stepNode__Prepend {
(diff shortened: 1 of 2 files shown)


kahboom (approved):
thank you @unsortedhashsets :)

PR #1692, opened 2023-04-25
feat(about): Add About modal to display front and back version

### LINKED ISSUE #696: Show version
As a user I would like to be able to check the version of Kaoto I am using.

lordrip:
### Context
At this moment, there's no mechanism in place to easily determine which version of Kaoto frontend or backend we're using.

This pull request provides a dedicated view to display those values.

### Changes
* Injects the KaotoUI version from the `package.json` file into a global variable `KAOTO_VERSION`.
* Wires up an endpoint from the backend to retrieve the backend version.
* Add an About modal to display the aforementioned versions.

### Example
* About menu item
image

* About modal
image

fix: [696](https://github.com/KaotoIO/kaoto-ui/issues/696)

### DIFF
--- a/src/store/settingsStore.test.tsx
+++ b/src/store/settingsStore.test.tsx
@@ -22,4 +22,15 @@ describe('settingsStore', () => {
     expect(result.current.settings.name).toEqual('Goji Berry');
     expect(result.current.settings.namespace).toEqual('KameletBinding');
   });
+
+  it('setBackendVersion', () => {
+    const { result } = renderHook(() => useSettingsStore());
+    expect(result.current.backendVersion).toEqual('');
+
+    act(() => {
+      result.current.setBackendVersion('1.0-test');
+    });
+
+    expect(result.current.backendVersion).toEqual('1.0-test');
+  });
 });
--- a/setupTests.ts
+++ b/setupTests.ts
@@ -4,6 +4,7 @@ import 'whatwg-fetch';
 import { stepsCatalog } from './src/stubs/steps';
 
 global.TextEncoder = TextEncoder;
+global.KAOTO_VERSION = '1.0-test';
 
 jest.mock('@kaoto/api', () => {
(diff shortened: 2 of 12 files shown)


src/store/settingsStore.test.tsx
```diff
@@ -22,4 +22,15 @@ describe('settingsStore', () => {
     expect(result.current.settings.name).toEqual('Goji Berry');
     expect(result.current.settings.namespace).toEqual('KameletBinding');
   });
+
+  it('setBackendVersion', () => {
+    const { result } = renderHook(() => useSettingsStore());
+    expect(result.current.backendVersion).toEqual('');
+
+    act(() => {
+      result.current.setBackendVersion('1.0-test');
```
unsortedhashsets, line 31: WDYT to use defined `global.KAOTO_VERSION = '1.0-test';` from `setupTests.ts` instead of hardcoded `result.current.setBackendVersion('1.0-test');` :thinking:

src/store/settingsStore.test.tsx
```diff
@@ -22,4 +22,15 @@ describe('settingsStore', () => {
...
+    });
+
+    expect(result.current.backendVersion).toEqual('1.0-test');
```
unsortedhashsets, line 34: WDYT to use defined `global.KAOTO_VERSION = '1.0-test';` from `setupTests.ts` instead of hardcoded `expect(result.current.backendVersion).toEqual('1.0-test');` :thinking:

src/store/settingsStore.test.tsx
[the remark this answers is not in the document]
unsortedhashsets: Answered upper

kahboom (approved):
Everything looks good to me, only thing is that I prefer the modal without the picture. That's probably not reason to block the merge but maybe for future review with @Mdenisco .

Issue #1708, opened 2023-04-27
Export route as png


apupier:
### Please describe the feature that you want to propose

it would be nice to export routes as png file.

1. ~As first iteration export all routes in the canvas.~ #1762
2. As a second might be nice to be able to export a single route at a time #1781 
3. As a third iteration to be able to select which route to export (but maybe not worthy) #1781

kahboom:
hey @lordrip 🙂  was this not already finished? probably best to assign yourself and update the status to prevent any duplication

kahboom:
ah nevermind i see there are more steps

lordrip:
😄 yes, I didn't want to close it as we have more iterations to do. That being said, I think that you raise a good point, if we keep the issue open, it will be open for the milestone, what if I extract the next iterations into other issues?

kahboom:
Hmm I wonder if this is big enough to be an epic

>what if I extract the next iterations into other issues?

I like that idea but I leave it to you, I think either way is fine. 😄

PR #1762, opened 2023-05-08
feat(viz): Export current flow to PNG

lordrip:
### Context
This commit adds the functionality to export the current canvas to a PNG file named after the integration name.

This iteration of exporting flows handles exporting the existing canvas to a PNG file. This functionality is placed under the kebab menu from the toolbar

### How to use
1. Open the kebab menu
image

2. Click on the `Export canvas to PNG` option
image
 
3. The PNG image is exported
image

### Notes
* The export functionality supports `light` and `dark` modes
* The exported image will contain what's currently visible on the canvas
* In case not all elements are visible on the screen, they won't appear on the canvas, another possibility is to try to center the canvas before exporting.

### Open questions
1. Would it be better to have a dedicated button on the canvas itself? for instance like
image
2. In case we want to keep the existing mechanism, would the following order be better?
image

Implements the first iteration of #1708

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "@rhoas/app-services-ui-shared": "^0.16.6",
     "ajv": "^8.12.0",
     "dagre": "^0.8.5",
+    "html-to-image": "^1.11.11",
     "immer": "^10.0.1",
     "lodash.isequal": "^4.5.0",
     "monaco-editor": "^0.38.0",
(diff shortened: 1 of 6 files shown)


igarashitm (approved):

kahboom (approved):

lordrip:
@igarashitm @kahboom do you have a preference between the different orders of the kebab menu?

| Current | Proposal |
| --- | --- |
| image | image |

PR #1764, opened 2023-05-08
fix: Provide prev/next step when querying step catalog to the backend

### LINKED ISSUE #1246: Add more context to mini-catalog calls.

When the mini-catalog opens, there is a call to the backend api like `http://....../api/v1/steps?dsl=KameletBinding&type=MIDDLE,END&namespace=default`

For https://github.com/KaotoIO/kaoto-backend/issues/224 I would like to have a bit more context. I would like to have the `id` of the previous and the following (if exists) steps to return a more accurate list of available steps.

I am going to add to the backend two more parameters to this endpoint https://kaotoio.github.io/kaoto-backend/api/index.html#api-StepResource-v1StepsGet

`previous-step` that should contain the `id` of the step immediately before the `+` button that opens the mini-catalog. If this is the first step of a branch, add the previous step too, the one that branches.
`following-step` that should contain the `id` of the step right after the `+` button, if such a step exists. If not, don't send it.

igarashitm:
Fixes: #1246


src/components/VisualizationStep.tsx
```diff
@@ -1,35 +1,29 @@
...
   const visualizationStore = useVisualizationStore();
-  const integrationJsonStore = useIntegrationJsonStore();
-  const visualizationService = new VisualizationService(integrationJsonStore, visualizationStore);
-  const stepsService = new StepsService(integrationJsonStore, nestedStepsStore, visualizationStore);
+  const visualizationService = new VisualizationService();
+  const stepsService = new StepsService();
   const showBranchesTab = VisualizationService.showBranchesTab(data.step);
   const showStepsTab = VisualizationService.showStepsTab(data);
   const supportsBranching = StepsService.supportsBranching(data.step);
+  const parentStepId = data.branchInfo?.parentStepUuid ?? undefined;
```
lordrip, line 26: [curious-question] shouldn't this be `undefined` if `branchInfo` is `undefined`? Maybe we don't need the `?? undefined` part

src/services/stepsService.ts
```diff
@@ -286,6 +286,15 @@ export class StepsService {
     return flattenMembers.concat(children.length ? StepsService.flattenSteps(children) : children);
   }
 
+  getFollowingStep(stepUuid?: string) {
+    if (!stepUuid) return undefined;
+    const currentStepIdx = this.findStepIdxWithUUID(stepUuid);
+    const steps = useIntegrationJsonStore.getState().integrationJson.steps;
+    return currentStepIdx !== -1 && steps.length > currentStepIdx + 1
```
lordrip, line 293: [side-comment] At the beginning I thought that we needed `steps.length >= currentStepIdx + 1` but after calculation, I think that it's ok 😄 

I went through this thought process:

Having this array: `[ 'a', 'b' ]`
**Case A:** Using `index=0` meaning `'a'`, for the condition `steps.length > currentStepIdx + 1` this would be `2 > 0 + 1` so yeah, 2 > 1.

**Case B:** Taking `'b'`, then it will be `2 > 1 + 1`, which is not `true` and it's ok since there's no additional step 🙆‍♂️

src/types/index.ts
```diff
@@ -173,6 +173,8 @@ export interface IStepQueryParams {
   namespace?: string;
   // e.g. 'START', 'END', 'MIDDLE'
   type?: string;
+  'previous-step'?: string;
+  'following-step'?: string;
```
lordrip, line 177: What are your thoughts about renaming these properties to the following? I think that would be easier to write a camel-cased version, there's an optional backend pull request if you want to do it: https://github.com/KaotoIO/kaoto-backend/pull/643

```suggestion
  'previousStep'?: string;
  'followingStep'?: string;
```

lordrip (dismissed):
LGTM.

PS: The differences between Import sorting and code formatting from IntelliJ and VSCode makes me cry 😭

src/components/VisualizationStep.tsx
igarashitm: fixned :+1:

src/types/index.ts
igarashitm: fixed :+1:

src/components/VisualizationStep.tsx
```diff
@@ -1,35 +1,31 @@
...
-  const visualizationService = new VisualizationService(integrationJsonStore, visualizationStore);
-  const stepsService = new StepsService(integrationJsonStore, nestedStepsStore, visualizationStore);
+  const visualizationService = new VisualizationService();
+  const stepsService = new StepsService();
   const showBranchesTab = VisualizationService.showBranchesTab(data.step);
   const showStepsTab = VisualizationService.showStepsTab(data);
   const supportsBranching = StepsService.supportsBranching(data.step);
+  const parentStepId = data.branchInfo?.parentStepUuid
+    ? stepsService.findStepWithUUID(data.branchInfo.parentStepUuid)?.id ?? undefined
+    : undefined;
```
igarashitm, line 28: @lordrip so I was using `UUID` before and getting 400 error. Now this gets the parent step `ID` successfully.

PR #1822, opened 2023-05-19
fix(stepsService): increment by 1 instead of 0.01

### LINKED ISSUE #1201: Spinner in config should increment by 1 instead of 0.01

In most cases, the values are integer and not decimals

spinnerIssue

tplevko:
With the new version of uniforms-patternfly - it's possible now to specify schema with step value - set by default to 1.

Closes https://github.com/KaotoIO/kaoto-ui/issues/1201

### DIFF
--- a/src/services/stepsService.ts
+++ b/src/services/stepsService.ts
@@ -649,6 +649,7 @@ export class StepsService {
       [label: string]: {
         type: string;
         defaultValue?: any;
+        step?: number;
         value?: any;
         description?: string;
         uniforms?: { placeholder: any };
@@ -663,6 +664,7 @@ export class StepsService {
         type,
         defaultValue,
         description,
+        ...(type === "number" || type === "integer") && {step : 1},
         uniforms: { placeholder: defaultValue },
         label: title,
       };


igarashitm (dismissed):

src/services/stepsService.ts
```diff
@@ -663,6 +664,7 @@ export class StepsService {
         type,
         defaultValue,
         description,
+        ...(type === "number" || "integer") && {step : 1},
```
kahboom, line 667: Hey @tplevko there is one minor issue with deepscan that points this out, https://deepscan.io/dashboard/#view=project&tid=19033&pid=22364&bid=658659&subview=pull-request&prid=1623867

So, best thing here is probably to write it like `...(type === "number" || type === "integer") && {step : 1},` instead of `...(type === "number" || "integer") && {step : 1},`.

kahboom (changes requested):
Minor point, otherwise everything looks great 👍

src/services/stepsService.ts
tplevko: Thanks for pointing that out - updated

lordrip (approved):

kahboom (approved):

Issue #801, opened 2022-10-12
Support multiple routes in the same file


Delawen:
### Context
Sometimes files contain more than one route. The front end right now is not considering a second route that may be available on the same file.

We may need a dropdown with the list of routes in the file. But then, the source code we see will not be the complete list of routes. Or would it?

### Roadmap
- [x] Create a new store that handles multiple flows (#1748)
- [x] Add flows identifiers to individual flows and steps (#1727)
- [x] Add initial support for multiple flows (#1749)
- [x] Support editing steps (#1770)
- [x] Wire up new backend endpoints (#1823)
- [x] Bug: Opening a similar step brings previous information (#1823)
- [x] Delete all flows (#1763)
- [X] Remove previous IntegrationJSONStore (#1835)
- [X] Create a new flow (#1870)
- [X] Flows interaction (#1880)
    * List all flows
    * Delete a flow
    * Being able to hide/collapse flows
- [X] Review remaining methods and ensure tests cover everything
    * #1901
    * #1903

### Associated bugs
- https://github.com/KaotoIO/kaoto-ui/issues/1911
- https://github.com/KaotoIO/kaoto-ui/issues/1910
- https://github.com/KaotoIO/kaoto-ui/issues/1896
- https://github.com/KaotoIO/kaoto-ui/issues/1894

### Examples
<details>
  <summary>Example of a single route</summary>

```yaml
- from:
    uri: "timer:yaml"
    parameters:
      period: "1000"
    steps:
      - set-body:
          constant: "Hello Camel K from yaml"
      - to: "log:info"
```
</details>

<details>
  <summary>Example of multiple routes</summary>

```yaml
- from:
    uri: "timer:yaml"
    parameters:
      period: "1000"
    steps:
      - set-body:
          constant: "Hello Camel K from yaml"
      - to: "log:info"
- from:
    uri: "timer:anotherone"
    parameters:
      period: "50000"
    steps:
      - set-body:
          constant: "Nobody expects the second route!"
      - to: "log:info"
```
</details>

brunoNetId:
I think this is a must.
it helps organised code and improve readability

rstroop:
Most of the Camel I write has multiple routes. I'd like to add that if we create directs it would be nice to have a dotted line between those routes showing their connection.

lordrip:
Related issue #1711

PR #1823, opened 2023-05-19
feat/wire-up new backend endpoints to support multiple flows

lordrip:
### Context
With the new endpoints introduced by https://github.com/KaotoIO/kaoto-backend/pull/638, we could start wiring up the source code editor and the related components to use the multiple flows approach.

### Changes
This pull request replaces the previous `v1/integrations` endpoint with the new `v2/integrations` endpoint that supports multiple flows.

All the changes were validated against the existing e2e tests to ensure that the previous functionality from the single flow is still in place.

### Known issues
At this stage, we cannot create new flows through the canvas, although writing multiple flows from the source code is possible.

#### SIngle flow
image

#### Multiple flows
image

Relates to: https://github.com/KaotoIO/kaoto-ui/issues/801


src/components/SettingsModal.tsx
```diff
@@ -51,39 +56,60 @@ export const SettingsModal = ({ handleCloseModal, isModalOpen }: ISettingsModal)
 
   const { addAlert } = useAlert() || {};
 
+  const onChangeDsl = useCallback((value: string) => {
```
igarashitm, line 59: Maybe in a separate one, but might be better to create a service and move this into it, like `FlowsService`

src/components/SourceCodeEditor.tsx
```diff
@@ -24,61 +25,65 @@ const SourceCodeEditor = (props: ISourceCodeEditor) => {
...
-
-  const fetchTheSourceCode = (integration: IIntegration, settings: ISettings) => {
-    const tmpIntegration = {
-      ...integration,
-      metadata: { ...integrationJson.metadata, ...settings },
-      dsl: settings.dsl.name,
+  }, [flows, properties]);
+
+  const fetchTheSourceCode = (currentFlowsWrapper: IFlowsWrapper, settings: ISettings) => {
+    const updatedFlowWrapper = {
```
igarashitm, line 55: And this one is also a candidate in `FlowsService`, then can we put all YAML>JSON and JSON>YAML work in a single service?

src/components/SourceCodeEditor.tsx
```diff
@@ -24,61 +25,65 @@ const SourceCodeEditor = (props: ISourceCodeEditor) => {
...
-  const previousJson = usePrevious(integrationJson);
+  const { flows, properties, setFlows } = useFlowsStore(({ flows, properties, setFlows }) => ({ flows, properties, setFlows }), shallow);
+  const previousFlows = usePrevious(flows);
+
   const schemaUri = settings.dsl.validationSchema
     ? process.env.KAOTO_API + settings.dsl.validationSchema
     : '';
 
   useEffect(() => {
-    let tmpInt: IIntegration = integrationJson;
```
igarashitm, line 35: And the content in this `useEffect` would be also a candidate to move out to a `FlowsService`

igarashitm (dismissed):
Approving - some routines in the components jumped out for me, looks like a needs of a service, but not directly related to this change

src/components/SourceCodeEditor.tsx
lordrip: Yup, that makes sense :ok_hand:

PR #1835, opened 2023-05-23
chore: Remove IntegrationJSONStore

lordrip:
After merging the last pull request about enabling the flowsStore, the next step is to remove the previous integrationJsonStore.

In addition to that, another needed step is to remove the temporary settings switch that toggles between the previous and the new store.

This pull request removes the integrationJsonStore and adjust the existing unit tests to comply with the new store.

relates to: https://github.com/KaotoIO/kaoto-ui/issues/801


igarashitm (approved):

src/components/AppendStepButton.tsx
```diff
@@ -58,7 +58,7 @@ export const AppendStepButton: FunctionComponent<IAddStepButton> = ({
             dsl: currentDSL,
             type: ValidationService.appendableStepTypes(step.type),
             previousStep: step.id,
-            followingStep: followingStep?.id,
+            followingStep: followingStepId,
```
kahboom, line 61: Tiny nitpick, do you mind if we use `nextStep` instead of `followingStep`? Only to stay consistent with `visualizationService`. Or we can change it there to `followingStep`.

kahboom (approved):
tiny nitpick but otherwise everything looks good 👍

src/components/AppendStepButton.tsx
lordrip: Which method of the `visualizationService`? I see that we're using `step` in the `stepsService` but I couldn't find a method with said name in `visualizationService`.

lordrip: mmm, now I get it, but in that sense, I believe that those aren't the same IDs. `nextStepUuid` refers to the UI-generated `UUID` and followingStepId refers to the backend-generated `ID`.

The latter is being used as a query parameter to give more context to the backend when retrieving the steps' list.

Could `nextStepId` be more appropriate? I mean changing it in the UI and the Backend as well.

PR #1836, opened 2023-05-23
Fix flaky code-editor-actions actions test

tplevko:
Found the source of instabilities for code-editor-actions tests in Chrome and Edge. The backend updates the yaml in editor on save - changing the indentation. This change is sometimes detected, sometimes not by the monaco editor, but only on these two browsers - didn't see this error in FF. I also didn't find  too much about the "undo" function. The alert is therefor a bit unpredictable, however the test was adjusted to verify the alert after the second "undo" invocation, where the alert fires regularly.

### DIFF
--- a/cypress/e2e/08-code-editor-actions/code_editor_actions.cy.js
+++ b/cypress/e2e/08-code-editor-actions/code_editor_actions.cy.js
@@ -88,14 +88,12 @@ describe('editing properties', () => {
     cy.get('[data-testid="viz-step-digitalocean"]').should('not.exist');
     cy.get('[data-testid="viz-step-set-header"]').should('not.exist');
 
-    // First click undo button => alert is displayed
+    // First click undo button => reverted automatic adjustments
     cy.editorClickUndoXTimes();
-
+    // Second click undo button => changes reverted & alert is displayed
+    cy.editorClickUndoXTimes();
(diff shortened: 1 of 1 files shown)


igarashitm (approved):

unsortedhashsets (approved):

lordrip (approved):
Thanks for investigating this @tplevko :muscle:

lordrip:
> I also didn't find too much about the "undo" function. The alert is therefor a bit unpredictable, however the test was adjusted to verify the alert after the second "undo" invocation, where the alert fires regularly.

Expanding a bit on what you're mentioning @tplevko, does this mean that the very first `undo` click doesn't work? If that's the case maybe there's a bug on the UI side then :thinking:

tplevko:
@lordrip actually yes - this is maybe however related to the way the editor handles the imported integration from file - e.g. the [EipAction.yaml](https://github.com/KaotoIO/kaoto-ui/blob/main/cypress/fixtures/EipAction.yaml) get's imported - is a valid yaml, but it actually contains indentination and some characters (e.g. double quotes instead of single quotes), which are actually replaced in the UI after the roundtrip to the backend, where the integration is parsed and the result is sent to the UI.

In this case - the first "undo" click really does nothing besides reverting the integration to the one which was imported.

PR #1869, opened 2023-05-29
fix(chromatic): Fix validation errors that shows in chromatic

lordrip:
### Context
Currently, the "chromatic" build step is tagging all commits as failed since there are a few errors showing up as now all endpoints are being mocked properly.

This pull request adds a small validation until we have a better-mocked setup to support all components and their stories.

It also removes the `SourceCodeEditor` story as we're not providing the necessary HTTP mocks to ensure that works properly.


Delawen (approved):

lordrip:
thanks @Delawen

PR #1870, opened 2023-05-29
Feat/add new flow

lordrip:
### Context
Currently, there is no mechanism to create a new flow in the canvas despite now the backend supporting such a feature.

This pull request adds the capability to create a new flow directly from the Toolbar.

### Changes:
* Remove the `Integration name` component from the `Toolbar`
* Add a new dropdown to select the new flow dsl
    * In case of the dsl of the existing flows is  the same as the new one, a new flow is created
    * On the contrary,  a modal is displayed to warn the user that the current flows will be lost
* In order to make the toolbar buttons more user-friendly, we're adding text to each of them, to convey their purpose easily.

#### Create a new flow
https://github.com/KaotoIO/kaoto-ui/assets/16512618/23c065ac-e599-4513-99c1-f96e4db38d87

#### Changing flow type
https://github.com/KaotoIO/kaoto-ui/assets/16512618/83cd5505-cb6e-4744-94f2-6405ef50d144

relates to: https://github.com/KaotoIO/kaoto-ui/issues/801

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -8,3 +8,7 @@ coverage:
         base: auto
 ignore:
   - "**/*.stories.tsx"
+  - "**/*.(test|stories).{ts,tsx}"
+  - "**/index.ts"
+  - "**/store/data/**"
+  - "**/stubs/**"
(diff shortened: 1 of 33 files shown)


igarashitm (approved):

igarashitm:
sonarcloud claims some "code smells", maybe that `\d` thing could be applied.

PR #1879, opened 2023-05-31
Add wait for API response before code sync

tplevko:
Add wait for API response before the code-sync is performed to the e2e tests.


lordrip:
@tplevko, do you think that it might be a good idea to introduce the "busy" concept as a matter to signal that the UI is doing something?

It might not be available for all circumstances and also we would need to check if impacts negatively the UI, but apart from that, I'm wondering if that would be helpful for the e2e.

The way that I imagine it, is that whenever you hit the sync button, there's an attribute set as "busy" and whenever the response is back, we switch it to "ready" or something like that. Maybe we could do something like that for the canvas as well.

tplevko:
In case it wouldn't negatively impact the UX, the "busy" concept could be a way to go. However for now I'd maybe leave as it is - the issues are only related to e2e tests, which are interacting with the UI "too quickly", causing the syncing issues with the backend, so I hope this could be solved with the waits for API responses.

unsortedhashsets (approved):

lordrip:
Sorry @tplevko, I didn't meant for you to do it :smile: , I meant if it could be helpful for the e2e  in the future, so I could start thinking on that and makes our lives easier and tested

Issue #1855, opened 2023-05-25
After end step removal - integration removing middle step instead


tplevko:
### Describe the Bug

This is a bug causing instabilities in the e2e test [step_actions_from_canvas.cy.js ](https://github.com/KaotoIO/kaoto-ui/blob/main/cypress/e2e/09-step_actions_from_canvas/step_actions_from_canvas.cy.js ), more specifically test named `Step detail - User selects a step, which opens the detail drawer`

Issue is with the integration - "timer -> choice -> log"

The issue occurs, when in the UI state after removing the last - "log" step. Instead of removing the ends step, the middle "choice" step gets removed and the "log" step remains in the integration.

See video.

Happens only occasionally and was only able to reproduce with the test. 

### Steps to Reproduce the Bug or Issue

1. run the test `Step detail - User selects a step, which opens the detail drawer`
2. try several times - mostly happens with headless browser

### Screenshots or Videos

https://github.com/KaotoIO/kaoto-ui/assets/4180208/015de431-ce0a-405d-99bc-ae6f261b965d

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [latest]

lordrip:
Thanks @tplevko for putting this together :+1:

PR #1921, opened 2023-06-06
Add basic set of multiflow tests

tplevko:
Add basic set of multiflow tests - both for canvas and editor.

The basic tests include:
Editor - based:

1. User deletes first route from multi-route using code editor
2. User deletes second route from multi-route using code editor
3. User deletes step from first route using code editor
4. User adds step to the first route using code editor
5. User reverts route deletion using code editor

Canvas - based:

1. User changes route type in the canvas
2. User shows and hides a route
3. User deletes routes in the canvas
4. User creates multiple routes in canvas

### DIFF
--- a/cypress/fixtures/IntegrationMultiFlow.yaml
+++ b/cypress/fixtures/IntegrationMultiFlow.yaml
@@ -1,0 +1,27 @@
+apiVersion: camel.apache.org/v1
+kind: Integration
+metadata:
+  name: ''
+spec:
+  flows:
+  - from:
+      uri: cron:cron
+      parameters:
+        schedule: '1000'
+      steps:
+      - set-body:
+          simple: body
+      - to:
+          uri: log:log1
+---
+apiVersion: camel.apache.org/v1
+kind: Integration
+metadata:
+  name: ''
+spec:
+  flows:
+  - from:
+      uri: timer:test
+      steps:
+      - to:
+          uri: log:log2
--- a/cypress/support/kaoto-ui-commands/default.js
+++ b/cypress/support/kaoto-ui-commands/default.js
@@ -1,95 +1,183 @@
 import 'cypress-file-upload';
 
 Cypress.Commands.add('openHomePage', () => {
-    let url = Cypress.config().baseUrl;
-    cy.visit(url);
-    cy.waitOpenHomePage();
+  let url = Cypress.config().baseUrl;
+  cy.visit(url);
+  cy.waitOpenHomePage();
 });
 
 Cypress.Commands.add('zoomOutXTimes', (times) => {
-    times = times ?? 1;
-    Array.from({ length: times }).forEach(() => {
-        cy.get('.react-flow__controls-button.react-flow__controls-zoomout').click()
-    });
+  times = times ?? 1;
+  Array.from({ length: times }).forEach(() => {
+    cy.get('.react-flow__controls-button.react-flow__controls-zoomout').click();
+  });
 });
 
 Cypress.Commands.add('waitVisualizationUpdate', () => {
-    cy.get('body').then((body) => {
-        /**
-         * If the code editor is visible, it means that we would need
-         * to wait for the getIntegration call, otherwise, the operation
-         * is synchronous
-         */
-        if (body.find('.code-editor').length > 0) {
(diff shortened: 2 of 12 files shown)


cypress/fixtures/IntegrationMultiFlow.yaml
```diff
@@ -0,0 +1,27 @@
...
+kind: Integration
+metadata:
+  name: ''
+spec:
+  flows:
+  - from:
+      uri: timer:test
+      steps:
+      - to:
+          uri: log:log2
```
lordrip, line 27: This integration type will be slightly different from what it is now. Later on, multiple routes from the `Integration` type will be handled inside of the `flows` array but this is in the works at the backend.

Can I suggest you another fixture in the meantime? This is a multiple Camel route fixture.

```yaml
- from:
    uri: cron:cron
    parameters:
      schedule: '1000'
    steps:
    - set-body:
        simple: body
    - to:
        uri: log:log1

- from:
    uri: timer:test
    steps:
    - to:
        uri: log:log2
```

cypress/support/kaoto-ui-commands/default.js
```diff
@@ -1,95 +1,183 @@
...
+ */
+Cypress.Commands.add('switchIntegrationType', (type) => {
+  cy.get('[data-testid="dsl-list-dropdown"]').click({ force: true });
+
+  cy.get('.pf-c-menu__item-text')
+    .contains(type)
+    .then((element) => {
+      cy.wrap(element).click();
+    });
+  cy.get('.pf-c-modal-box__footer').children('.pf-c-button.pf-m-primary').click();
```
lordrip, line 109: Would it be convenient to have a better handler for the `Accept` button? I can add it :smiley:

cypress/support/kaoto-ui-commands/steps.js
```diff
@@ -64,7 +64,7 @@ Cypress.Commands.add('deleteStep', (step, stepIndex) => {
         const previousStepsCount = elem.length;
 
         /** Delete the given step */
-        cy.get(`[data-testid="viz-step-${step}"]`).eq(stepIndex).trigger('mouseover').children('[data-testid="configurationTab__deleteBtn"]').click({ force: true });
+        cy.get(`[data-testid="viz-step-${step}"]`,{withinSubject:null}).eq(stepIndex).trigger('mouseover').children('[data-testid="configurationTab__deleteBtn"]').click({ force: true });
```
lordrip, line 67: Is this `{withinSubject: null}` ok or would be better to have a dedicated handler from the UI side?

lordrip (dismissed):
LGTM, thanks Tomas :muscle:

cypress/fixtures/IntegrationMultiFlow.yaml
tplevko: sure, thanks for letting me know - will convert this to the `flows` once it's in the backend. Also the second fixture - I'll extend the tests and add there also camelRoute type tests, so the covered area represents all the possible route options.

cypress/support/kaoto-ui-commands/steps.js
tplevko: found this is probably redundant - removed, thanks for noticing this

cypress/support/kaoto-ui-commands/default.js
tplevko: should be ok now

cypress/support/kaoto-ui-commands/steps.js
```diff
@@ -80,6 +80,7 @@ Cypress.Commands.add('openStepConfigurationTab', (step, EIP, stepIndex) => {
     if (!EIP) {
         cy.get('[data-testid="kaoto-right-drawer"]').should('be.visible');
     }
+    cy.waitVisualizationUpdate();
```
tplevko, line 83: This should be workarround for https://github.com/KaotoIO/kaoto-ui/issues/1855

lordrip: Awesome, I'll link your message on that issue, right now is just saying that the issue got mentioned on this pull request and I think it will come in handy to ensure that the bug was fixed.

lordrip (approved):

PR #1927, opened 2023-06-07
Fix - sync connector yaml without need to open code editor

### LINKED ISSUE #1361: Adding steps while having open the Code Editor makes it flash
### Describe the Bug

When opening the Code Editor, if the user add steps, it flashes while being updated.

1. Start with an empty canvas
2. Open the Code Editor
3. Add a step
4. Notice how the Code Editor flashes between updates

### Screenshots or Videos

[Screencast from 2023-03-03 10-10-28.webm](https://user-images.githubusercontent.com/16512618/222679658-d9a76182-59d6-461e-bd05-0fe5a3f5537c.webm)

### Platform

- OS: [Linux]
- Browser: [Firefox]
- Version: [0.7.1-dev]

### LINKED ISSUE #1368: Opening the Code Editor creates a flash at the top
### Describe the Bug

When opening the Code Editor, there's a content flash at the top, and for the first time only, it shows a message of "Start editing / Drag and drop a file or upload one."

1. Start with a clean canvas
2. Open the Code Editor
3. Notice the flash of "Start editing / Drag and drop a file or upload one."
4. Close the Code Editor and open it again
5. Notice the "Any invalid code will be replaced after sync."

### Screenshots or Videos

[Screencast from 2023-03-03 10-38-38.webm](https://user-images.githubusercontent.com/16512618/222686750-b6350315-1f90-4ef0-8e46-924963739bba.webm)

### Platform

- OS: [Linux]
- Browser: [Firefox]
- Version: [0.7.1-dev]

### LINKED ISSUE #1796: Can't deploy an integration without previously opening the code editor in the UI
### Describe the Bug

When users try to create an integration without opening the "Code Editor" tab - the Post request to the backend only sends the integration stub (see first image).
Only after opening the "Code Editor" tab and trying again to deploy the integration the Post request contains the complete integration resource. 

1. Create integration (e.g. timer to log)
2. don't open the editor tab
3. try to deploy the integration to the connected cluster
4. see the post request only contains the integration stub

### Screenshots or Videos

Screenshot from 2023-05-15 14-54-54

Screenshot from 2023-05-15 14-55-28

### Platform

- Version: [latest]

tplevko:
Fixes https://github.com/KaotoIO/kaoto-ui/issues/1796
Fixes https://github.com/KaotoIO/kaoto-ui/issues/1361
Fixes https://github.com/KaotoIO/kaoto-ui/issues/1368
The code editor doesn't need to be opened for the code sync to be performed prior to integration deployment to Kubernetes cluster.

### DIFF
--- a/src/components/SourceCodeEditor.tsx
+++ b/src/components/SourceCodeEditor.tsx
@@ -1,12 +1,11 @@
-import { usePrevious } from '../hooks';
 import { StepErrorBoundary } from './StepErrorBoundary';
-import { fetchCapabilities, fetchIntegrationJson, fetchIntegrationSourceCode } from '@kaoto/api';
+import { fetchIntegrationJson } from '@kaoto/api';
 import { useFlowsStore, useIntegrationSourceStore, useSettingsStore } from '@kaoto/store';
-import { CodeEditorMode, ICapabilities, IFlowsWrapper, ISettings } from '@kaoto/types';
+import { CodeEditorMode } from '@kaoto/types';
 import { CodeEditor, CodeEditorControl, Language } from '@patternfly/react-code-editor';
 import { Alert } from '@patternfly/react-core';
 import { CheckCircleIcon, EraserIcon, RedoIcon, UndoIcon } from '@patternfly/react-icons';
-import { useEffect, useRef, useState } from 'react';
+import { useRef } from 'react';
 import { EditorDidMount } from 'react-monaco-editor';
 import { useDebouncedCallback } from 'use-debounce';
 import { shallow } from 'zustand/shallow';
@@ -24,57 +23,13 @@ interface ISourceCodeEditor {
 export const SourceCodeEditor = (props: ISourceCodeEditor) => {
   const editorRef = useRef<Parameters<EditorDidMount>[0] | null>(null);
   const { sourceCode, setSourceCode } = useIntegrationSourceStore();
-  const [syncedCode, setSyncedCode] = useState('');
+  const syncedSourceCode = useIntegrationSourceStore((state) => state.syncedSourceCode, shallow);
   const { settings, setSettings } = useSettingsStore();
-  const { flows, properties, metadata, setFlowsWrapper } = useFlowsStore(
-    ({ flows, properties, metadata, setFlowsWrapper }) => ({
-      flows,
-      properties,
-      metadata,
-      setFlowsWrapper,
-    }),
-    shallow,
-  );
-  const previousFlows = usePrevious(flows);
+  const setFlowsWrapper = useFlowsStore((state) => state.setFlowsWrapper, shallow);
 
   const schemaUri = settings.dsl.validationSchema
     ? process.env.KAOTO_API + settings.dsl.validationSchema
     : '';
-
-  useEffect(() => {
-    if (previousFlows === flows || !Array.isArray(flows[0]?.steps)) return;
-
-    if (flows[0].dsl !== null && flows[0].dsl !== settings.dsl.name) {
-      fetchCapabilities().then((capabilities: ICapabilities) => {
-        capabilities.dsls.forEach((dsl) => {
-          if (dsl.name === flows[0].dsl) {
-            const tmpSettings = { ...settings, dsl };
-            setSettings(tmpSettings);
-            fetchTheSourceCode({ flows, properties, metadata }, tmpSettings);
-          }
-        });
-      });
-    } else {
-      fetchTheSourceCode({ flows, properties, metadata }, settings);
-    }
-  }, [flows, properties]);
-
-  const fetchTheSourceCode = (currentFlowsWrapper: IFlowsWrapper, settings: ISettings) => {
-    const updatedFlowWrapper = {
-      ...currentFlowsWrapper,
-      flows: currentFlowsWrapper.flows.map((flow) => ({
-        ...flow,
-        dsl: settings.dsl.name,
-      })),
-    };
-
-    fetchIntegrationSourceCode(updatedFlowWrapper, settings.namespace).then((newSrc) => {
-      if (typeof newSrc === 'string') {
-        setSourceCode(newSrc);
-        setSyncedCode(newSrc);
-      }
-    });
-  };
 
   /**
    * On detected changes to YAML state, issue POST to external endpoint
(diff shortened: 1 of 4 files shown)


igarashitm:
Cool good catch overall - wondering if we can put synced code into existing `integrationSourceStore` instead of introducing dedicated store. We might even want to introduce such as `SourceSyncService` and move the source sync logic into it, so that all YAML>JSON and JSON>YAML logic are put together there.

tplevko:
Thank you for the suggestion @igarashitm! I moved the synced code handling to integrationSourceStore. Regarding the `SourceSyncService` - sounds like outside the scope of this PR - but we can definitely add this later

src/components/Visualization.tsx
```diff
@@ -51,10 +53,55 @@ const Visualization = () => {
     shallow,
   );
 
-  const flows = useFlowsStore((state) => state.flows);
+  const { settings, setSettings } = useSettingsStore((state) => state);
+  const { setSourceCode } = useIntegrationSourceStore((state) => state);
```
igarashitm, line 57: Now this could be like
```suggestion
  const { setSourceCode, setSyncedSourceCode } = useIntegrationSourceStore((state) => state);
```
@lordrip is there any benefit to do even like this?
```typescript
  const { setSourceCode, setSyncedSourceCode } = useIntegrationSourceStore(({setSourceCode, setSyncedSourceCode) => ({setSourceCode, setSyncedSourceCode}));
```

tplevko: thanks - updated

lordrip: > @lordrip is there any benefit to do even like this?

From the functional point of view, both should be the same, what would be different though is to use the `shallow` function when using the store, because otherwise this component will be rendered any time that a property changes, despite being relevant or not, so my recommendation would be, bring the least amount of data required, and do the composition in the function itself, so save some calculations.

```typescript
/**
* IMHO good as we only use what we need and the `shallow`
* function operates over 2 properties instead of the whole `state` object
*/
import { shallow } from 'zustand';

const { prop1, prop2 } = useMyStore((state) => ({
    prop1: state.prop1,
    prop2: state.prop1 / 1_000, // making calculations
}) , shallow);
```

```typescript
/**
* Same as before, but now we're adding overhead by checking whether
* the object or its properties changed
*/
import { shallow } from 'zustand';

const { prop1, prop2 } = useMyStore((state) => state, shallow);
```

```typescript
/**
* Definitively not good as we render anytime that something change
*/

const { prop1, prop2 } = useMyStore();
```

lordrip:
I tried out the branch and it works well, I'm just wondering what would be the purpose of the new property.

igarashitm:
@lordrip IIRC `syncedSourceCode` was already there, it's "reformatted" code by the backend after YAML>JSON and JSON>YAML, altough I also wonder if it's really needed separately. Maybe reafactor in the another attempt?

src/components/SourceCodeEditor.tsx
```diff
@@ -24,58 +23,14 @@ interface ISourceCodeEditor {
...
-    ({ flows, properties, metadata, setFlowsWrapper }) => ({
-      flows,
-      properties,
-      metadata,
-      setFlowsWrapper,
-    }),
-    shallow,
-  );
-  const previousFlows = usePrevious(flows);
+  const { setFlowsWrapper } = useFlowsStore(({ setFlowsWrapper }) => ({ setFlowsWrapper, }), shallow,);
```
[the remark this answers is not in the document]
tplevko: thanks - updated

src/components/SourceCodeEditor.tsx
```diff
@@ -24,58 +23,14 @@ interface ISourceCodeEditor {
 export const SourceCodeEditor = (props: ISourceCodeEditor) => {
   const editorRef = useRef<Parameters<EditorDidMount>[0] | null>(null);
   const { sourceCode, setSourceCode } = useIntegrationSourceStore();
-  const [syncedCode, setSyncedCode] = useState('');
+  const { syncedSourceCode } = useIntegrationSourceStore((state) => state);
```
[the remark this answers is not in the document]
tplevko: thanks - updated

lordrip:
> `syncedSourceCode`

Yes, it was there as an internal state of `sourceCodeEditor`, so now I'm wondering if this is still needed :thinking: but is not blocking the pull request by any means

igarashitm (dismissed):
LGTM - however I agree having 2 separate source code in the store smells, there might be a better way. @tplevko can you file an issue for investigating potential refactoring?

tplevko:
> can you file an issue for investigating potential refactoring?

Sure, just opened a separate refactor issue for investigation and consolidation of these two states - https://github.com/KaotoIO/kaoto-ui/issues/1959

src/components/Visualization.tsx
```diff
@@ -51,10 +53,54 @@ const Visualization = () => {
     shallow,
   );
 
-  const flows = useFlowsStore((state) => state.flows);
+  const { settings, setSettings } = useSettingsStore((state) => state);
+  const { setSourceCode, setSyncedSourceCode } = useIntegrationSourceStore((state) => state);
```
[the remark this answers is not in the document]
lordrip: oh, is not a stupid question by any means :smiley:, actually, you're right, it shouldn't be necessary at all. Let me check what the `shallow` function does in case of a `function`.

src/components/Visualization.tsx
```diff
@@ -51,10 +53,55 @@ const Visualization = () => {
...
+      metadata,
+    }), shallow
+  );
   const visualizationService = useMemo(() => new VisualizationService(), []);
   const stepsService = useMemo(() => new StepsService(), []);
 
+  const previousFlows = usePrevious(flows);
+
+  useEffect(() => {
+    if (previousFlows === flows || !Array.isArray(flows[0].steps)) return;
```
lordrip, line 72: This will cause an error whenever we delete all existing flows by trying to access `undefined`. [It happened to me in the past](https://github.com/KaotoIO/kaoto-ui/pull/1927/files#diff-adf7a8db55b4b1624ee2eaf3c331f474481bde7d8e79927dfec164216372bb99L45) :sweat_smile:.

```suggestion
    if (previousFlows === flows || !Array.isArray(flows[0]?.steps)) return;
```

lordrip (dismissed):
LGTM, thanks for taking care, and sorry for the delayed review. Notice that there's a suggestion open to avoid an error whenever we delete all flows, but apart from that, everything looks good :ok_man:

lordrip (approved):

src/components/Visualization.tsx
tplevko: thanks for review - updated

PR #1950, opened 2023-06-12
Make search input width static

### LINKED ISSUE #1245: Steps catalog search bar split into 2 lines in Mozilla Firefox
### Describe the Bug

There's an inconsistent behavior between Google Chrome and Mozilla Firefox in regard to the Steps catalog search bar that causes the search bar to be split into two lines

This happens when searching for a step.

#### Mozilla Firefox - no search
Screenshot from 2023-02-13 10-36-50

#### Mozilla Firefox - search
Screenshot from 2023-02-13 10-36-58

#### Google Chome - no search
Screenshot from 2023-02-13 10-35-55

#### Google Chome - search
Screenshot from 2023-02-13 10-36-13

1. Open the steps catalog
2. Write a letter to start searching
3. Notice how the input field gets wider in Chrome or the search bar is split in Firefox

### Screenshots or Videos

_No response_

### Platform

- OS: [Linux]
- Browser: [Chrome, Firefox]
- Version: [0.6.2-dev]

tplevko:
Fix for https://github.com/KaotoIO/kaoto-ui/issues/1245

Changed the input search to have static width - not causing the "start, actions, end" toggles to change position in the toolbar. The static width causes in FF not to overflow the search bar and doesn't cause moving the toggles to second row. See recordings:

Previous state:

Peek 2023-06-12 11-09

With changes:

Peek 2023-06-12 11-08

fixes: #1245

### DIFF
--- a/src/components/Catalog.tsx
+++ b/src/components/Catalog.tsx
@@ -106,7 +106,7 @@ export const Catalog = ({ handleClose }: { handleClose: () => void }) => {
             <>
               <ToolbarItem style={{ padding: '0', marginRight: '0' }}>
                 <InputGroup>
-                  <SearchInput
+                  <SearchInput style={{ width: '250px' }}
                     name={'stepSearch'}
                     id={'stepSearch'}
                     type={'search'}


src/components/Catalog.tsx
```diff
@@ -106,17 +106,19 @@ export const Catalog = ({ handleClose }: { handleClose: () => void }) => {
...
-                    id={'stepSearch'}
-                    type={'search'}
-                    placeholder={'search for a step...'}
-                    aria-label={'search for a step'}
-                    value={query}
-                    onChange={(_e, s) => changeSearch(s)}
-                    onClear={() => setQuery('')}
-                    ref={searchInputRef}
-                  />
+                  <div className={'catalog__searchInputWrapper'}>
```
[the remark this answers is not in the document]
tplevko: TBH I tried to figure out how to do it the way you are considering, but in this case the behaviour - meaning the "expanding" search field in case I typed some input remained unchanged, so I chose this approach.

src/components/Catalog.tsx
```diff
@@ -106,7 +106,7 @@ export const Catalog = ({ handleClose }: { handleClose: () => void }) => {
             <>
               <ToolbarItem style={{ padding: '0', marginRight: '0' }}>
                 <InputGroup>
-                  <SearchInput
+                  <SearchInput style={{ width: '250px' }}
```
lordrip, line 109: In general, this should be avoided, since it will generate a "new" change every time we define a new object on each render. Probably here is not a big deal, though.

Another way to approach this situation might be leveraging `flexbox` and letting the field fill the available space at all times.

lordrip (approved):
Thanks for taking care of this @tplevko :smiley:

src/components/Catalog.tsx
tplevko: sure, thanks for the hint - will have a look how to do this better

PR #1967, opened 2023-06-14
feat(viz): Provide an empty state for the canvas

### LINKED ISSUE #1909: Indicate no routes are created

In case there are no routes created - the UI should let the user know. 
Maybe two places would be suitable - either in the canvas, or in the routes dropdown (currently there is just empty list). 

Screenshot from 2023-06-05 15-32-23

lordrip:
### Context
Currently, whenever there's no visible route in the canvas, a full gray area is shown.

image

### Changes
The goal for this commit is to provide more information to the user about the status of the routes in the canvas itself. When there are no visible routes, an according message will be displayed.

#### No visible routes
image

#### No routes
image

fixes: https://github.com/KaotoIO/kaoto-ui/issues/1909


lhein:
looks good!

igarashitm (dismissed):

src/components/Visualization.tsx
```diff
@@ -243,28 +261,32 @@ const Visualization = () => {
...
-            deleteKeyCode={null}
-            zoomOnDoubleClick={false}
-            className="panelCustom"
-          >
-            {/*<MiniMap nodeBorderRadius={2} className={'visualization__minimap'} />*/}
-            <VisualizationControls />
-            <Background color="#aaa" gap={16} />
-          </ReactFlow>
-        </div>
+        {visibleFlowsInformation.isCanvasEmpty ? (
```
lordrip, line 264: OMG, I forgot to add tests for this clause, I'll add this test case and merge it afterward.

@tplevko covering this condition at the integration test side would be enough or should we add an e2e as well?

lordrip: Here we have [the missing test](https://github.com/KaotoIO/kaoto-ui/pull/1967/files#diff-86c560251a1c2da7c959dcf452624515f2e2ca08ddfbced394c24f58425b0a07R79)

tplevko: Well from my POV, I think the test you added should be enough, can add more testing for this in e2e if requested.

PR #2009, opened 2023-06-21
Fix e2e tests after backend quarkus update

mkralik3:
CORS needs to be set after Quarkus3 update on backend https://github.com/KaotoIO/kaoto-backend/pull/695  
[More info](https://github.com/KaotoIO/kaoto-backend#cors)

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -57,6 +57,8 @@ jobs:
         ports:
           # will assign a random free host port
           - 8081/tcp
+        env:
+          QUARKUS_HTTP_CORS_ORIGINS: "http://localhost:1337"
     steps:
       - name: 👷‍♀️ Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


igarashitm (approved):
Although these 2 are still failing, I think it's a separate problem. This fix improves the situation for sure. Actually these 2 are failing also locally for me.
* 04-source_code/source_code.cy.js 
* 08-code-editor-actions/code_editor_actions.cy.js

mkralik3:
If they were working before the Quarkus3 upgrade, maybe, there is another issue that backend tests didn't catch. I think we can merge this and investigate failing tests afterwards

cc: @tplevko

lordrip:
Either that or is not deployed yet :sob: 
image

mkralik3:
this PR doesn't resolve the nightly cluster, only e2e tests in UI test suite. 
Also, I have asked mmelko and the nightly cluster was deployed manually and there is a trigger to refetch new backend image, so the deployment with ENV needs to be update manually there.

PR #2034, opened 2023-06-23
fix(vscode): Routes gets reset upon opening in vscode.

### LINKED ISSUE #285: Routes are reset to empty when opening them with the Kaoto Editor (main branch of kaoto UI)

lordrip:
### Context
Currently, there are two effects taking care of the syncing process between the VS Code file and KaotoUI (and source code) in the `KogitoEditorIntegrationProvider.tsx` file. Said effects react when one of the several dependencies changes, like flows or settings.

This causes a recursive call since the same effects are updating the setting object from which they depend.

### Changes
The fix is for the effect to rely only on the intended property.

This is still colliding with the Sourcecode sync that it's happening in the `Visualization.tsx` component. A follow-up will be needed to consolidate both scenarios.

fixes: https://github.com/KaotoIO/vscode-kaoto/issues/285

### DIFF
--- a/src/components/Visualization.tsx
+++ b/src/components/Visualization.tsx
@@ -97,7 +97,7 @@ const Visualization = () => {
       fetchCapabilities().then((capabilities: ICapabilities) => {
         capabilities.dsls.forEach((dsl) => {
           if (dsl.name === flows[0].dsl) {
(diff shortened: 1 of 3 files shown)


apupier:
Tested locally in VS Code and I do not reproduce anymore the initial issue

PR #2035, opened 2023-06-23
fix(vscode): Missing backend version in vscode

### LINKED ISSUE #287: Backend version is missing from about dialog (main branch of Kaoto UI)
image

lordrip:
### Context
Currently, there's a race condition caused between the `KogitoEditorIntegrationProvider.tsx` and the `KaotoToolbar.tsx` as the latter holds an outdated reference to the settings object, causing the backend version to be removed.

### Changes
The fix is to use the correct API for the 'setSettings()' hook in which we only pass the property to be updated and nothing more.

Also, there's an icon property set in the `settingsStore` which causes the comparison to be slower than needed, as it holds the base64 representation of an image.

fixes: https://github.com/KaotoIO/vscode-kaoto/issues/287

### DIFF
--- a/src/components/KaotoToolbar.tsx
+++ b/src/components/KaotoToolbar.tsx
@@ -86,9 +86,8 @@ export const KaotoToolbar = ({
   // fetch default namespace from the API,
   useEffect(() => {
     fetchDefaultNamespace().then((data) => {
-      const newSettings = settings;
-      newSettings.namespace = data.namespace;
-      setSettings(newSettings);
+      const namespace = data.namespace;
+      setSettings({ namespace});
     });
   }, []);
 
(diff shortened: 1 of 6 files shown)


src/components/KaotoToolbar.tsx
```diff
@@ -86,9 +86,8 @@ export const KaotoToolbar = ({
   // fetch default namespace from the API,
   useEffect(() => {
     fetchDefaultNamespace().then((data) => {
-      const newSettings = settings;
```
lordrip, line 89: This was the main issue, in this hook, the `settings` object is outdated.

apupier:
works in VS code:

image

PR #2064, opened 2023-06-28
fix(flows): New route button should use supportsMultipleFlows property

### LINKED ISSUE #2055: "New route" button should not add multiple CRDs
### Describe the Bug

Currently the "New route" button adds multiple CRDs for `Integration`, `Kamelet` and `KameletBinding`. Since multiple CRDs require bigger consideration for having multiple contexts, we decided to
* Integration kind: Add routes in the existing CRD
* Kamelet kind: Disable `New route`
* KameletBinding: Disable `New route`

Until we fully support multiple contexts.

1. Choose `Integration`, `Kamelet` or `KameletBinding`
2. Click `New route` button

### Screenshots or Videos

_No response_

### Platform

- Version: [e.g. 0.30.0]

lordrip:
### Context
Currently, there's a `supportsMultipleFlows` property from the capabilities endpoint from the backend that the UI is not leveraging yet.

This property is meant to signal the UI whether a DSL supports multiple flows or not.

### Changes
This pull request makes use of the said property to disable the New route button in case the UI is dealing with those kinds of DSL.

At the moment of writing this code, those DSLs are `Kamelet` and `KameletBindings`

#### for Kamelet and KameletBindings, add new routes were allowed from the UI
image

#### Kamelet and KameletBindings options are disabled if a route of the same DSL already exists
image

fixes: https://github.com/KaotoIO/kaoto-ui/issues/2055


apupier (dismissed):
- Works fine for me locally.
- route ids not in sync between UI and source code but doubt it will be related to this PR, Ricardo is aware and will check
- Note that in fact Kaoto backend doesn't support multiple routes correctly currently for Integration Kind (contrarily to the attribute value claimed by the API), see https://github.com/KaotoIO/kaoto-backend/issues/733 and https://github.com/KaotoIO/kaoto-backend/issues/726

PR #2075, opened 2023-06-30
chore(e2e): Fix broken e2e

lordrip:
### Context
With the introduction of https://github.com/KaotoIO/kaoto-backend/pull/734, the multiple flows support for Integration CRDs is no longer an individual CRD per flow but rather a single CRD with many flows in it.

### Changes
This commit takes that into consideration by updating the fixtures and the associated lines.

In addition to that, the editorDeleteLine was refactored to leverage the monaco GoToLine functionality to consistently reach a given line.

### DIFF
--- a/cypress/support/kaoto-ui-commands/editor.js
+++ b/cypress/support/kaoto-ui-commands/editor.js
@@ -29,16 +29,32 @@ Cypress.Commands.add('editorAddText', (line, text) => {
 
 Cypress.Commands.add('editorDeleteLine', (line, repeatCount) => {
     repeatCount = repeatCount ?? 1;
+    // Open the Go to Line dialog
     cy.get('.code-editor')
         .click()
         .type(
-            '{pageUp}' +
-            '{downArrow}'.repeat(line) +
-            '{shift}' +
-            '{downArrow}'.repeat(repeatCount) +
-            '{backspace}',
-            { delay: 1 }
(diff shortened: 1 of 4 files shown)


cypress/support/kaoto-ui-commands/editor.js
```diff
@@ -29,16 +29,32 @@ Cypress.Commands.add('editorAddText', (line, text) => {
 
 Cypress.Commands.add('editorDeleteLine', (line, repeatCount) => {
     repeatCount = repeatCount ?? 1;
+    // Open the Go to Line dialog
```
lordrip, line 32: Jump directly to a given line.

cypress/support/kaoto-ui-commands/editor.js
```diff
@@ -29,16 +29,32 @@ Cypress.Commands.add('editorAddText', (line, text) => {
...
+    // Type the line number to delete
+    cy.get('input[aria-describedby="quickInput_message"][aria-controls="quickInput_list"]')
+        .click()
+        .type(
+            `${line + 1}` +
+            '{enter}',
+            { delay: 1 },
+        );
+
+    // Delete the line as many times as specified
```
lordrip, line 50: This will count as separate interactions, meaning that if we want to `undo` or `redo` we need to do it as many times as we deleted lines.

PR #2110, opened 2023-07-07
chore: reformat files and enable lint check

lordrip:
### Context
Currently, there are 2 scripts in the `package.json` to help with code formatting but none of them are enforced.

### Changes
* Reformat all files in the `src` folder
* Enable a GitHub action to enforce the checks

### Notes
* Prerequisite of https://github.com/KaotoIO/kaoto-ui/pull/2103
* There are still 16 warnings left regarding React dependencies:

✖ 16 problems (0 errors, 16 warnings)


.github/workflows/format.yml
```diff
@@ -0,0 +1,21 @@
...
+    name: 💅 Format and Linter checks
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: 🤹‍♀️ Install modules
+        run: yarn
+      - name: 💅 Run format check
+        run: yarn format
+      - name: 💅 Run lint check
+        run: yarn lint
```
igarashitm, line 21: Can `yarn format (prettier)` and `yarn lint (eslint)` co-exist? i.e. can we do `yarn format:fix` and `yarn lint:fix` and satisfy both? I think we don't have to, rather we can remove `yarn lint`.
Also, in order to make this workflow "required" for PR check, we need higher permission which I don't have... the easier alternative is to add `yarn format` into existing required workflow, such as unit-tests.yaml instead.

lordrip: Yup, there might be rules that could collide, but it seems that we don't have any at the moment.

I also added the workflow as required now.

PR #2115, opened 2023-07-10
fix(source): Deleting last route doesn't sync the code

### LINKED ISSUE #2086: Deleting last route from the canvas doesn't get reflected in the code editor
### Describe the Bug

When users delete the route using the delete dropdown - the route sources stay in the code editor.

1. create a route
2. specify route steps
3. delete the route using the dropdown
4. see the route source remains in the code editor

### Screenshots or Videos

Peek 2023-07-03 09-38

### Platform

- Version: [last]

lordrip:
### Context
Currently, there's a check in the `Visualization.tsx` component that ensures that at least one flow exists in the `FlowsStore`.

This check prevented to sync of the source code when all flows were deleted.

### Changes
The fix is to sync the source code whenever there's a change in the Flows array. In addition to that, the current DSL needs to be updated [whenever a new flow is added](https://github.com/KaotoIO/kaoto-ui/pull/2115/files#diff-a16b0f9b3604fc7b64098ed9eb1ff564b5333d8a6a7656a2f650c245d60144d1R35).

fixes: https://github.com/KaotoIO/kaoto-ui/issues/2086

### DIFF
--- a/src/components/DSL/NewFlow.tsx
+++ b/src/components/DSL/NewFlow.tsx
@@ -32,6 +32,7 @@ export const NewFlow: FunctionComponent<PropsWithChildren> = () => {
          * we don't need to do anything special, just add a new flow if
          * supported
          */
+        setSettings({ dsl, namespace });
         addNewFlow(dsl.name);
       } else {
         /**
@@ -42,7 +43,7 @@ export const NewFlow: FunctionComponent<PropsWithChildren> = () => {
         setIsConfirmationModalOpen(true);
       }
     },
-    [addNewFlow],
+    [addNewFlow, namespace, setSettings],
   );
 
   return (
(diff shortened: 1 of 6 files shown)


src/components/DSL/NewFlow.tsx
```diff
@@ -32,6 +32,7 @@ export const NewFlow: FunctionComponent<PropsWithChildren> = () => {
          * we don't need to do anything special, just add a new flow if
          * supported
          */
+        setSettings({ dsl, namespace });
```
lordrip, line 35: We need to update the current DSL whenever we add a new flow, since in this code path when [there are no flows, it assumes that the DSL is the same](https://github.com/KaotoIO/kaoto-ui/pull/2115/files#diff-a16b0f9b3604fc7b64098ed9eb1ff564b5333d8a6a7656a2f650c245d60144d1R27), because `Array.every` returns `true` for empty arrays.

```
[].every(() => true);
// prints `true`
```

PR #2158, opened 2023-07-17
chore: Turn project into monorepo and add a sandbox application

### LINKED ISSUE #2159: Provide a sandbox app to simulate vscode and multiplying architecture
### Context
At the moment of writing, in order to test the UI functionality in `vscode-kaoto`, we need to build both packages locally and link them through `yarn link`.

Although this works, is not ideal, since adds additional steps when developing a new feature or fixing a bug.

### Goal
The goal for this issue is to provide a web application that also leverages the multiplying architecture, similar to this one: https://github.com/KaotoIO/kaoto-ui/pull/1566, and reuse the existing e2e tests against it.

lordrip:
### Context
In the context of https://github.com/KaotoIO/kaoto-ui/discussions/1759, and with the aim to improve the `vscode-kaoto` testing experience by creating a sandbox application that leverages the multiplier architecture, this commit turns the project into a `monorepo`.

This will bring a few challenges around prefixing yarn commands with the right package, but at the same time, it will allow us to bring other packages in an easy way.

### Notes
There are a few things that change when turning the repository into a monorepo, one of those things is how we interact with the packages.

previously, we could issue a command like the following to run the `e2e` tests:
```bash
yarn run e2e:headless
```

now, with this pull request, it would be:
```bash
yarn workspace @kaoto/kaoto-ui run e2e:headless
```

or can be the previous command if we enter the correct directory first:
```
cd kaoto-dir/packages/kaoto-ui

[user@host kaoto-ui]$ yarn run e2e:headless
```

The same goes for linking the `vscode-kaoto` repository with `kaoto-ui`, now the command would be:
```
[user@host vscode-kaoto-folder]$ yarn link kaoto-ui-dir/packages/kaoto-ui
```

Another thing that changed is the build output, Previously, the resulting artifacts were located at `/dist` now they will be at `/packages/kaoto-ui/dist`, This means that we need to adapt those workflows that checkout `kaoto-ui` repository and build it (`kaoto-standalone`, `kaoto-backend`)

### Pending topics
- [X] Fix `dockerfile` build stage: The issue here is that we need to copy into the container the correct files and build them
- [X] Fix E2E tests: Locally they are passing, but on CI we need to check the paths
- [X] Deepscan: Since all `/src` files were moved, all alerts are being reset.
- [X] SonarCloud Code Analysis: All alerts are being reset since all `/src` files were moved.
- [ ] Adjust e2e to be able to look up inside iframes when testing @kaoto/sandbox

### Web version
image

### Adapting a simple e2e
[Screencast from 2023-07-18 06-05-37.webm](https://github.com/KaotoIO/kaoto-ui/assets/16512618/8a2a9e27-a57f-4bbf-a73b-11c9266792c8)

### e2e + sandbox application
This is being worked on, from e2e tests, 50% are passing without many adjustments, the remaining tests need deep verification to determine whether the failure is related to an actual problem showcased because of the sandbox application or that the test needs adjustments.

Fixes: https://github.com/KaotoIO/kaoto-ui/issues/2159


packages/kaoto-ui/cypress/e2e/00-iframe-test/about.cy.js
```diff
@@ -0,0 +1,32 @@
+describe('Settings: About', () => {
```
lordrip, line 1: Adapted from [packages/kaoto-ui/cypress/e2e/01-settings/about.cy.js](https://github.com/KaotoIO/kaoto-ui/blob/7c803defcea2b6ce95b615b806aba55e86c1f0dc/packages/kaoto-ui/cypress/e2e/01-settings/about.cy.js)

lordrip: The change is to replace `.get()` functions with `getIframeBody()` calls when applies. I'm wondering if this can be inferred at runtime, meaning that we will use the same function no matter if we're running the test against `Kaoto UI` or the sandbox.

packages/kaoto-ui/cypress/support/kaoto-ui-commands/default.js
```diff
@@ -1,5 +1,22 @@
 import 'cypress-file-upload';
 
+Cypress.Commands.add('getIframeBody', () => {
```
lordrip, line 3: @tplevko here we have the function that I created for searching the iframe, I believe that we could introduce a check inside of the body to execute the `.get()` against the `iframe` or not.

packages/kaoto-ui/cypress/support/kaoto-ui-commands/default.js
```diff
@@ -1,5 +1,27 @@
 import 'cypress-file-upload';
 
+Cypress.Commands.add('isomorphicGet', (selector) => {
```
lordrip, line 3: @tplevko here we have the function that I created for searching the iframe. It leverages an environment variable `IFRAME_ENABLED` to look for the multiplying architecture's iframe or look directly over the root document.

packages/kaoto-ui/cypress/support/kaoto-ui-commands/steps.js
```diff
@@ -21,79 +21,79 @@ Cypress.Commands.add('dragAndDropFromCatalog', (source, target, catalog, targetI
 
 Cypress.Commands.add('replaceEmptyStepMiniCatalog', (step, stepIndex) => {
     stepIndex = stepIndex ?? 0;
-    cy.get('[data-testid="viz-step-slot"]').eq(stepIndex).click();
+    cy.isomorphicGet('[data-testid="viz-step-slot"]').eq(stepIndex).click().click();
```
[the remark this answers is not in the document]
tplevko: This might be causing also the other failures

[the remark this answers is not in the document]
lordrip: Thanks for checking this out @tplevko , I updated the pull request :)

